PR #4, opened 2023-04-07
Controller

aojea:
This has it working, it creates a loop that monitors for the existing clusters, and start a service lb controller per cluster.
The containers created for the lb has a special label, and when the cloud-provider-kind is stopped, they are deleted


aojea:
can I do yolo?

/lgtm
/approve

PR #7, opened 2023-04-18
Add config flag to instruction running cpk

mikejoh:
👋🏻 I updated the docs a bit here (a bit of tidying up), I also added the `--config` flag in the instruction to start a `kind` cluster. I referenced the provided `kind.yaml`  file that holds all customization needed to run the cluster with an external cloud provider. I think that makes sense in this case, but I'm not sure, shall that be another section in the instructions perhaps?


mikejoh:
I'll close this, needs more work basically!

PR #8, opened 2023-04-21
feat: support ipv6

czybjtu:


Makefile
```diff
@@ -12,7 +12,7 @@ export GOROOT GO111MODULE CGO_ENABLED
 
 
 build:
-	go build -v -o "$(OUT_DIR)/$(KIND_CLOUD_BINARY_NAME)" $(KIND_CLOUD_BUILD_FLAGS) cmd/main.go
+	go build -o "$(OUT_DIR)/$(KIND_CLOUD_BINARY_NAME)" $(KIND_CLOUD_BUILD_FLAGS) cmd/main.go
```
BenTheElder, line 15: Unrelated change?

czybjtu: yes. I think it's unnecessary to print so much compilation infomation. I can checkout it if you think it's not appropriate.

BenTheElder: At the very least this should be in a different commit.

I think it's useful to print this, personally.

czybjtu: Have been checkouted.

aojea:
Thanks, have you tested it?

czybjtu:
Yes. I have tested the following scenarios:
- IPv4 single stack
- IPv6 single stack
- dualstack

aojea:
/lgtm
/approve

TODO: add tests

aojea:
Thanks very much

PR #40, opened 2024-04-02
reject connections if backends are down

aojea:


pkg/loadbalancer/proxy.go
```diff
@@ -61,6 +61,8 @@ defaults
 frontend {{$index}}-frontend
   bind {{ $data.BindAddress }}
   default_backend {{$index}}-backend
+  # reject connections if all backends are down
+  tcp-request connection reject if { nbsrv({{$index}}-backend) lt 1 }
```
BenTheElder, line 65: https://www.haproxy.com/documentation/haproxy-configuration-manual/latest/#4-tcp-request%20connection

BenTheElder: so this is basically https://stackoverflow.com/a/17712250 :-)

aojea: is the only test failing in CI :) https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/124117/pull-kubernetes-e2e-kind-cloud-provider-loadbalancer/1774932103805276160

aojea:
/hold cancel

the existing e2e test with bash is very flaky, I'm going to remove it too

pkg/loadbalancer/proxy.go
BenTheElder: that's awesome!

Issue #41, opened 2024-04-04
Question about macOS / Windows support


gary-archer:
Ideally I'd like a solution where I can demo some fairly advanced Kubernetes setups related to APIs in the same way on Linux / macOS / Windows (development computer) hosts. In future this might include multi-cluster setups to simulate multi-region systems. Right now though I switch to port mapping on macOS / Windows, which is quite limited.

In the README you mention the following:

```text
Only tested with docker and Linux, but podman and Windows and Macusers feedback will be helpful to support those platforms.
```

The haproxy container runs within a kind (bridge) network which I believe is not reachable on macOS / Windows due to Docker limitations.

```bash
docker ps --format '{{ .Image }} {{ json .Networks }}'

kindest/haproxy:v20230606-42a2262b "kind"
```

I have disabled all firewalls and I think my problem is a failure to connect to the haproxy container:

```bash
kind delete cluster 2>/dev/null
kind create cluster
kubectl label node kind-control-plane node.kubernetes.io/exclude-from-external-load-balancers-
kubectl -n applications  apply -f lb-service-local.yaml

EXTERNAL_IP=$(kubectl -n applications get svc/lb-service-local -o jsonpath="{.status.loadBalancer.ingress[0].ip}")
curl "http://$EXTERNAL_IP/hostname"

[R](curl: (28) Failed to connect to 172.19.0.4 port 80 after 21029 ms: Timed out)
```

I am no networking expert so I thought I'd ask for a hint or two on how you see a macOS or Windows solution working. Is there anything other than port mapping you think will work?  E.g. would switching to podman on those platforms make the networking work the same as on a Linux host? Or are there any other steps you would recommend?

It would be great if I could get this working in a cross-platform way and I am happy to help with any testing. Regards.

BenTheElder:
It's possible to setup something like https://github.com/chipmk/docker-mac-net-connect

Alternatively: run your tests from another container on the docker bridge. It will have no problem accessing the loadbalancer etc.

aojea:
I have a branch with a possible implementation, I need beta testers though

gary-archer:
Great - let me know details and I will help out with the beta testing. I work at a company that does Identity and Access Management and we want to demonstrate cloud native setups, e.g:

- Providing a web app code example that connects to APIs in the local cluster
- But without asking developers who run it to do anything too low level on their machines

[...]

PR #42, opened 2024-04-05
Loadbalancer for Mac

aojea:
This is a pretty simple solution with some limitations, it uses the existing port forwarding capabilities of docker to expose the loadbalancer ports on the host, and then adds the same IP of the container to the loopback address and start proxying from user space the service ports on the service address to the portmapped ports.

Limitations:
- Service port changes are not updated as docker does not dynamically update forwarded ports
- Kind binary needs permissions to add IP address to interfaces and to listen on privileged ports
- Overlapping IP on the hosts

TODO
- [x] check cleanup works
- [ ] handle overlapping ips or at least warn

Fixes: #41


aojea:
I´m going to merge because this seems to work nice now

PR #45, opened 2024-04-18
Allow installation with go install

skitt:
Moving main.go to the root of the project allows installation using

    go install sigs.k8s.io/cloud-provider-kind@latest

with a correctly-named result. (Installation currently works with .../cmd, but the resulting binary is $GOBIN/cmd.)

(This avoids relying on release artefacts for installation — they haven’t been pubished for 0.0.2 :wink:)


aojea:
@skitt thanks for this change

/lgtm
/approve

BenTheElder:
thanks!

PR #56, opened 2024-04-23
Enable docker compose

electrocucaracha:
Utilizing `docker-compose` tool can reduce the number of instructions used to deploy the cloud provider service.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,11 +49,18 @@ Another alternative is to run it as a container, but this will require to mount
 the docker socket inside the container:
 
 ```sh
-docker build . -t aojea/cloud-provider-kind:v0.1
+docker build . -t cloud-provider-kind
 # using the host network
-docker run --rm --network host -v /var/run/docker.sock:/var/run/docker.sock aojea/cloud-provider-kind:v0.1
+docker run --rm --network host -v /var/run/docker.sock:/var/run/docker.sock cloud-provider-kind
 # or the kind network
-docker run --rm --network kind  -v /var/run/docker.sock:/var/run/docker.sock aojea/cloud-provider-kind:v0.1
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -49,11 +49,8 @@ Another alternative is to run it as a container, but this will require to mount
 the docker socket inside the container:
 
 ```sh
-docker build . -t aojea/cloud-provider-kind:v0.1
-# using the host network
-docker run --rm --network host -v /var/run/docker.sock:/var/run/docker.sock aojea/cloud-provider-kind:v0.1
-# or the kind network
-docker run --rm --network kind  -v /var/run/docker.sock:/var/run/docker.sock aojea/cloud-provider-kind:v0.1
+# using the `kind` network (`host` is the default value for NET_MODE)
+NET_MODE=kind docker compose up -d
```
BenTheElder, line 53: I feel like we should document how to do this without compose still.

E.G. if we start to publish a container image that is pre-built (when the project is more mature) then you will want to skip building most of the time and we'll want this simple docker run step again (or binary run on the host)

We're also going to want to figure out e.g. podman, where you won't be mounting the socket necessarily.

aojea:
hi Victor, thanks for the PR but as Ben say we have a considerable use base of podman users ... and now nerdctl

I would rather prefer to make the deployment containerized ... but in a standard way that serves all ...  but I don't know if this will work also in Mac and Windows, since we need access to the host

aojea:
I don't mind adding it as an additional subsection or alternative way, but I prefer to keep the existing ... we may also rename the iamge to not use my github and use something more generic

aojea:
/lgtm
/approve

Gracias Victor

Issue #58, opened 2024-04-24
Port Protocol UDP Support


skizot722:
Hello! First of all, let me say amazing work on this project. I've set it up without issue and tested for some sample TCP services and it worked flawlessly.  My actual use case for a local testing environment involves a UDP service port, and I figured out after a bit of digging that UDP is currently [not supported](https://github.com/kubernetes-sigs/cloud-provider-kind/blob/main/pkg/loadbalancer/proxy.go#L109). Are there any plans to support that in the near future?  What are the limitations to allowing UDP right now in the proxy?

aojea:
we are using haproxy as implementation of loadbalancer because we already use it in KIND, and it does not support UDP https://github.com/haproxy/haproxy/issues/62 

We need to understand first if there are other LB we can use to replace haproxy, like envoy ... but that may take time ... not very top priority :(

skizot722:
Thanks @aojea - that's an _interesting_ thread.

danwinship:
The haproxy thread there suggests that nginx does UDP proxying for some value of UDP proxying, and [it looks like ingress-nginx supports UDP Ingress](https://github.com/kubernetes/ingress-nginx/blob/main/docs/user-guide/exposing-tcp-udp-services.md), which suggests nginx's UDP support is good enough for what Kubernetes Services need.

aojea:
back when we discussed this in kind , several years ago , I've tried with envoy since also offers an API to control remotely
and accepts UDP https://www.envoyproxy.io/docs/envoy/latest/start/sandboxes/udp ... it is also used widely for cloud providers as loadbalancers so it will provide the most realistic implementation 

If I have time I will try to replace haproxy with envoy ... unless someone beats me 😄

aojea:
I've also added an example https://github.com/kubernetes-sigs/cloud-provider-kind/blob/main/examples/loadbalancer_udp_tcp.yaml so you can try it

skizot722:
@aojea - wow, thank you so much for the work and lightning fast turnaround!

Issue #57, opened 2024-04-24
[Flake]  LoadBalancers should not have connectivity disruption during rolling update with externalTrafficPolicy=Local [Slow] 


aojea:
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cloud-provider-kind/56/pull-cloud-provider-kind-conformance-parallel-ga-only/1782890037956841472

> I0423 22:11:53.246754 74697 loadbalancer.go:1791] Number of running nodes: 2, number of updated ready pods: 0 in daemonset test-lb-rolling-update
I0423 22:11:54.420971 74697 loadbalancer.go:1725] Got error testing for reachability of http://172.18.0.5:80/echo?msg=hello: Get "http://172.18.0.5:80/echo?msg=hello": write tcp 172.18.0.1:47414->172.18.0.5:80: write: connection reset by peer
I0423 22:11:54.421265 74697 loadbalancer.go:1725] Got error testing for reachability of http://172.18.0.5:80/echo?msg=hello: Get "http://172.18.0.5:80/echo?msg=hello": read tcp 172.18.0.1:47422->172.18.0.5:80: read: connection reset by peer
I0423 22:11:54.421564 74697 loadbalancer.go:1725] Got error testing for reachability of http://172.18.0.5:80/echo?msg=hello: Get "http://172.18.0.5:80/echo?msg=hello": read tcp 172.18.0.1:47430->172.18.0.5:80: read: connection reset by peer
I0423 22:11:54.421898 74697 loadbalancer.go:1725] Got error testing for reachability of http://172.18.0.5:80/echo?msg=hello: Get "http://172.18.0.5:80/echo?msg=hello": read tcp 172.18.0.1:47432->172.18.0.5:80: read: connection reset

```
{ failed [FAILED] Encountered too many errors when doing HTTP requests to the load balancer address. Success rate is 90.51%, and the minimum allowed threshold is 95.00%.

```

PR #61, opened 2024-05-03
replace haproxy with envoy (static configuration)

aojea:
Switching to envoy will give us more features and also the capacity to configure remotely the loadbalancers

Fixes: #58, #57

### DIFF
--- a/pkg/loadbalancer/proxy.go
+++ b/pkg/loadbalancer/proxy.go
@@ -235,7 +235,7 @@ func proxyUpdateLoadBalancer(ctx context.Context, clusterName string, service *v
 		if !container.IsRunning(name) {
 			// if is flapping
 			if checks > 0 {
-				return false, fmt.Errorf("container %s is not stable")
+				return false, fmt.Errorf("loadbalancer %s is not stable", name)
 			}
 			return false, nil
 		}
--- a/pkg/loadbalancer/proxy.go
+++ b/pkg/loadbalancer/proxy.go
@@ -89,12 +89,14 @@ static_resources:
   - name: cluster_{{$index}}
     connect_timeout: 0.25s
     type: STATIC
-    lb_policy: ROUND_ROBIN
+    lb_policy: RANDOM
     health_checks:
       - timeout: 1s
         interval: 2s
         unhealthy_threshold: 3
         healthy_threshold: 1
+        always_log_health_check_failures: true
+        always_log_health_check_success: true
         http_health_check:
           path: /healthz
     load_assignment:
--- a/pkg/loadbalancer/proxy.go
+++ b/pkg/loadbalancer/proxy.go
@@ -23,11 +23,11 @@ const proxyConfigPath = "/etc/envoy/envoy.yaml"
 
 // proxyConfigData is supplied to the loadbalancer config template
 type proxyConfigData struct {
-	HealthCheckPort int             // is the same for all ServicePorts
-	ServicePorts    map[string]data // key is the IP family and Port to support MultiPort services
+	HealthCheckPort int                    // is the same for all ServicePorts
+	ServicePorts    map[string]servicePort // key is the IP family and Port and Protocol to support MultiPort services
 }
 
-type data struct {
+type servicePort struct {
 	// frontend
 	Listener endpoint
 	// backend
@@ -48,14 +48,14 @@ admin:
 
 static_resources:
   listeners:
-  {{- range $index, $data := .ServicePorts }}
+  {{- range $index, $servicePort := .ServicePorts }}
   - name: listener_{{$index}}
     address:
       socket_address:
-        address: {{ $data.Listener.Address }}
-        port_value: {{ $data.Listener.Port }}
-        protocol: {{ $data.Listener.Protocol }}
-    {{- if eq $data.Listener.Protocol "UDP"}}
+        address: {{ $servicePort.Listener.Address }}
+        port_value: {{ $servicePort.Listener.Port }}
+        protocol: {{ $servicePort.Listener.Protocol }}
+    {{- if eq $servicePort.Listener.Protocol "UDP"}}
     udp_listener_config:
       downstream_socket_config:
         max_rx_datagram_size: 9000
@@ -85,7 +85,7 @@ static_resources:
   {{- end }}
 
   clusters:
-  {{- range $index, $data := .ServicePorts }}
+  {{- range $index, $servicePort := .ServicePorts }}
   - name: cluster_{{$index}}
     connect_timeout: 0.25s
     type: STATIC
@@ -102,7 +102,7 @@ static_resources:
     load_assignment:
       cluster_name: cluster_{{$index}}
       endpoints:
-	  {{- range $address := $data.Cluster }}
+	  {{- range $address := $servicePort.Cluster }}
         - lb_endpoints:
           - endpoint:
               health_check_config:
@@ -144,7 +144,7 @@ func generateConfig(service *v1.Service, nodes []*v1.Node) *proxyConfigData {
 		HealthCheckPort: hcPort,
 	}
 
-	servicePortConfig := map[string]data{}
+	servicePortConfig := map[string]servicePort{}
 	for _, ipFamily := range service.Spec.IPFamilies {
 		// TODO: support UDP
 		for _, port := range service.Spec.Ports {
@@ -175,7 +175,7 @@ func generateConfig(service *v1.Service, nodes []*v1.Node) *proxyConfigData {
 				}
 			}
 
-			servicePortConfig[key] = data{
+			servicePortConfig[key] = servicePort{
 				Listener: endpoint{Address: bind, Port: int(port.Port), Protocol: string(port.Protocol)},
 				Cluster:  backends,
 			}
--- a/pkg/loadbalancer/proxy.go
+++ b/pkg/loadbalancer/proxy.go
@@ -72,6 +72,10 @@ static_resources:
               typed_config:
                 '@type': type.googleapis.com/envoy.extensions.filters.udp.udp_proxy.v3.Route
                 cluster: cluster_{{$index}}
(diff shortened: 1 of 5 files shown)


pkg/loadbalancer/proxy.go
```diff
@@ -31,45 +29,57 @@ type proxyConfigData struct {
 
 type data struct {
 	// frontend
-	BindAddress string // *:Port for IPv4 :::Port for IPv6
+	Listener endpoint
 	// backend
-	Backends map[string]string // key: node name  value: IP:Port
+	Cluster []endpoint // key: node name  value: IP:Port
```
danwinship, line 34: the comment no longer applies

pkg/loadbalancer/proxy.go
```diff
@@ -31,45 +29,57 @@ type proxyConfigData struct {
...
+          typed_config:
+            "@type": type.googleapis.com/envoy.extensions.filters.network.tcp_proxy.v3.TcpProxy
+            stat_prefix: destination
+            cluster: cluster_{{$index}}
+  {{- end }}
+
+  clusters:
+  {{- range $index, $data := .ServicePorts }}
+  - name: cluster_{{$index}}
+    connect_timeout: 0.25s
```
danwinship, line 90: presumably this is not the default value, in which case what's wrong with the default and why do we need to change it? (needs a comment?)

pkg/loadbalancer/proxy.go
```diff
@@ -31,45 +29,57 @@ type proxyConfigData struct {
 
 type data struct {
```
danwinship, line 30: your type name is bad and you should feel bad. :slightly_smiling_face:

pkg/loadbalancer/proxy_test.go
```diff
@@ -181,53 +197,17 @@ func Test_proxyConfig(t *testing.T) {
...
+						Listener: endpoint{Address: "0.0.0.0", Port: 443},
+						Cluster:  []endpoint{{"192.168.8.2", 31497}, {"192.168.8.3", 31497}},
+					},
 				},
 			},
 			wantConfig: `
-global
-log /dev/log local0
-log /dev/log local1 notice
-daemon
```
danwinship, line 198: the fact that `wantConfig` is now `""` seems wrong

pkg/loadbalancer/proxy.go
```diff
@@ -50,14 +51,37 @@ static_resources:
...
     address:
-      socket_address: { address: {{ $data.Listener.Address }}, port_value: {{ $data.Listener.Port }} }
+      socket_address:
+        address: {{ $data.Listener.Address }}
+        port_value: {{ $data.Listener.Port }}
+        protocol: {{ $data.Listener.Protocol }}
+    {{- if eq $data.Listener.Protocol "UDP"}}
+    udp_listener_config:
+      downstream_socket_config:
+        max_rx_datagram_size: 9000
```
danwinship, line 61: again, not clear why this needs to be specified?

pkg/loadbalancer/proxy.go
```diff
@@ -50,14 +51,37 @@ static_resources:
...
+        stat_prefix: cluster_{{$index}}
+        matcher:
+          on_no_match:
+            action:
+              name: route
+              typed_config:
+                '@type': type.googleapis.com/envoy.extensions.filters.udp.udp_proxy.v3.Route
+                cluster: cluster_{{$index}}
+        upstream_socket_config:
+          max_rx_datagram_size: 9000
```
danwinship, line 75: it seems weird that the UDP section is so different from the TCP section but I assume it would make sense if I knew more about envoy config...

pkg/loadbalancer/proxy.go
```diff
@@ -115,11 +139,11 @@ func generateConfig(service *v1.Service, nodes []*v1.Node) *proxyConfigData {
 	for _, ipFamily := range service.Spec.IPFamilies {
 		// TODO: support UDP
```
danwinship, line 147: kill that comment

pkg/loadbalancer/proxy.go
```diff
@@ -115,11 +139,11 @@ func generateConfig(service *v1.Service, nodes []*v1.Node) *proxyConfigData {
...
 		for _, port := range service.Spec.Ports {
-			if port.Protocol != v1.ProtocolTCP {
+			if port.Protocol != v1.ProtocolTCP && port.Protocol != v1.ProtocolUDP {
 				klog.Infof("service port protocol %s not supported", port.Protocol)
 				continue
 			}
-			key := fmt.Sprintf("%s_%d", string(ipFamily), port.Port)
+			key := fmt.Sprintf("%s_%d_%s", string(ipFamily), port.Port, port.Protocol)
```
danwinship, line 153: (It's inconsistent to cast `ipFamily` to `string` but not `port.Protocol`. You don't need to cast either of them.)

pkg/loadbalancer/proxy_test.go
```diff
@@ -76,9 +76,9 @@ func Test_generateConfig(t *testing.T) {
 			want: &proxyConfigData{
 				HealthCheckPort: 32000,
 				ServicePorts: map[string]data{
-					"IPv4_80": data{
-						Listener: endpoint{Address: "0.0.0.0", Port: 80},
-						Cluster:  []endpoint{{"10.0.0.1", 30000}, {"10.0.0.2", 30000}},
+					"IPv4_80_TCP": data{
+						Listener: endpoint{Address: "0.0.0.0", Port: 80, Protocol: string(v1.ProtocolTCP)},
```
danwinship, line 80: You could just say `Protocol: "TCP"`; the templates already depend on the constants having specific expected values

pkg/loadbalancer/proxy.go
```diff
@@ -90,6 +90,13 @@ static_resources:
     connect_timeout: 0.25s
     type: STATIC
     lb_policy: ROUND_ROBIN
+    health_checks:
+      - timeout: 1s
+        interval: 2s
+        unhealthy_threshold: 3
+        healthy_threshold: 1
```
danwinship, line 97: again would be good to explain why these particular numbers

danwinship (commented):
I don't really know much about haproxy or envoy, but...

pkg/loadbalancer/proxy.go
aojea: it is horrible , isn't it ? the problem is that I run out of ideas and end with the same names in all the indirection layers   😄

pkg/loadbalancer/proxy.go
aojea: copy paste from the totally obscure envou config world

pkg/loadbalancer/proxy.go
aojea: how do you find these things :)

pkg/loadbalancer/proxy.go
aojea: copy paste

aojea:
> Kubernetes e2e suite: [It] [sig-network] LoadBalancers should not have connectivity disruption during rolling update with externalTrafficPolicy=Cluster [Slow] 

it is failing in all jobs that means most probably the healtchecker is not working

pkg/loadbalancer/proxy.go
danwinship: maybe at least 

`// FIXME: figure out what these numbers mean and if we should change them or make them configurable`

danwinship: (though fwiw the "real" cloud providers have tons of random hardcoded constants in their health check configs too)

pkg/loadbalancer/proxy_test.go
aojea: fixed

aojea:
ok, found the bug, it was on the configuration , envoy has some hot restart mechanism that capture the HUP signal and does not allow to get the new config

Alos tested UDP and it works fine

pkg/loadbalancer/proxy.go
aojea: this is explained here https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/health_check.proto.html

I finally understood the API docs, is just that all of them are defined in proto and they render this way on the docs and at least to me was hard to find the values and the meaning

pkg/loadbalancer/proxy_test.go
aojea: lazy copy and paste

pkg/loadbalancer/proxy.go
aojea: It seems there are some internal details that make it different https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/udp_filters/udp_proxy , is interesting how complex must be to support UDP in the load balancing worlds, we can see haproxy does not want to implement it

pkg/loadbalancer/proxy.go
aojea: I prefer to not expose low level options like this to the users until we have a super stable version that works in CI and we have some strong use cases

pkg/loadbalancer/proxy.go
aojea: this value is wrong, is too agressive, I ve restored it to the default value that is 5 seconds, the time envoy waits for the request to finish

pkg/loadbalancer/proxy.go
aojea: https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/udp_filters/udp_proxy#example-configuration

pkg/loadbalancer/proxy.go
aojea: replaced

aojea:
And affinity working too ... added also examples and improve the resilience ... nice experience to learn the envoy APIs , super powerful and expressive though the documentation is super complex

aojea:
green, I want to merge to unblock the CI testing, this way we can move on

PR #63, opened 2024-05-04
Set correct IPMode in LoadBalancerStatus

danwinship:
cloud-provider-kind proxies all traffic rather than redirecting it, so we need to set `IPMode: Proxy` to indicate that (since `IPMode: VIP` is otherwise assumed if `IP` is set rather than `Hostname`).

(IPMode field is beta in 1.30.)

/assign @aojea

### DIFF
--- a/pkg/loadbalancer/server.go
+++ b/pkg/loadbalancer/server.go
@@ -14,6 +14,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	cloudprovider "k8s.io/cloud-provider"
 	"k8s.io/klog/v2"
+	"k8s.io/utils/ptr"
 	"sigs.k8s.io/cloud-provider-kind/pkg/constants"
 	"sigs.k8s.io/cloud-provider-kind/pkg/container"
 )
@@ -65,10 +66,18 @@ func (s *Server) GetLoadBalancer(ctx context.Context, clusterName string, servic
 		}
 	}
 	if ipv4 != "" && svcIPv4 {
-		status.Ingress = append(status.Ingress, v1.LoadBalancerIngress{IP: ipv4, Ports: portStatus})
+		status.Ingress = append(status.Ingress, v1.LoadBalancerIngress{
+			IP:     ipv4,
+			IPMode: ptr.To(v1.LoadBalancerIPModeProxy),
+			Ports:  portStatus,
+		})
 	}
 	if ipv6 != "" && svcIPv6 {
-		status.Ingress = append(status.Ingress, v1.LoadBalancerIngress{IP: ipv6, Ports: portStatus})
+		status.Ingress = append(status.Ingress, v1.LoadBalancerIngress{
+			IP:     ipv6,
+			IPMode: ptr.To(v1.LoadBalancerIPModeProxy),
+			Ports:  portStatus,
+		})
 	}
 
 	return status, true, nil


aojea:
Lgtm once we validate this works.against old clusters 

The test failed is about load balancer source range filtering, I don't remember a.similar.error, was this test running before?

aojea:
Following up the discussion  from https://github.com/kubernetes-sigs/cloud-provider-kind/pull/64#discussion_r1590265825 here, my quick test failed with non existing fields, but despite that, if we are going to use this for testing we'll end in the same situation we had in kind, so passing the kubernetes version of the cluster sounds like something we'll have to do sooner or later , @danwinship can you add something like this?

```diff
diff --git a/pkg/controller/controller.go b/pkg/controller/controller.go
index 6163aab..cbb0465 100644
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -85,8 +85,14 @@ func (c *Controller) Run(ctx context.Context) {
                                continue
                        }
 
+                       ver, err := kubeClient.ServerVersion()
+                       if err != nil {
+                               klog.Errorf("Failed to to get cluster version %s: %v", cluster, err)
+                               continue
+                       }
+
                        klog.V(2).Infof("Creating new cloud provider for cluster %s", cluster)
-                       cloud := provider.New(cluster, c.kind)
+                       cloud := provider.New(cluster, c.kind, ver)
                        ccm, err := startCloudControllerManager(ctx, cluster, kubeClient, cloud)
                        if err != nil {
                                klog.Errorf("Failed to start cloud controller for cluster %s: %v", cluster, err)
@@ -112,7 +118,7 @@ func (c *Controller) Run(ctx context.Context) {
 // getKubeClient returns a kubeclient depending if the ccm runs inside a container
 // inside the same docker network that the kind cluster or run externally in the host
 // It tries first to connect to the external endpoint
-func (c *Controller) getKubeClient(ctx context.Context, cluster string) (kubernetes.Interface, error) {
+func (c *Controller) getKubeClient(ctx context.Context, cluster string) (*kubernetes.Clientset, error) {
        httpClient := &http.Client{
                Timeout: 5 * time.Second,
                Transport: &http.Transport{
diff --git a/pkg/loadbalancer/server.go b/pkg/loadbalancer/server.go
index e096c1b..b34d385 100644
--- a/pkg/loadbalancer/server.go
+++ b/pkg/loadbalancer/server.go
@@ -12,6 +12,7 @@ import (
 
        v1 "k8s.io/api/core/v1"
        metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+       "k8s.io/apimachinery/pkg/version"
        cloudprovider "k8s.io/cloud-provider"
        "k8s.io/klog/v2"
        "sigs.k8s.io/cloud-provider-kind/pkg/constants"
@@ -19,13 +20,16 @@ import (
 )
 
 type Server struct {
+       k8sVersion    *version.Info // used to keep compatibility between new features and old clusters
        tunnelManager *tunnelManager
 }
 
 var _ cloudprovider.LoadBalancer = &Server{}
 
-func NewServer() cloudprovider.LoadBalancer {
-       s := &Server{}
+func NewServer(ver *version.Info) cloudprovider.LoadBalancer {
+       s := &Server{
+               k8sVersion: ver,
+       }
        if runtime.GOOS == "darwin" || runtime.GOOS == "windows" {
                s.tunnelManager = NewTunnelManager()
        }
@@ -64,6 +68,7 @@ func (s *Server) GetLoadBalancer(ctx context.Context, clusterName string, servic
                        svcIPv6 = true
                }
        }
+       // TODO dan KEP-1860 , only add ipMode if cluster GreaterOrEqual 1.30.0
        if ipv4 != "" && svcIPv4 {
                status.Ingress = append(status.Ingress, v1.LoadBalancerIngress{IP: ipv4, Ports: portStatus})
        }
diff --git a/pkg/provider/cloud.go b/pkg/provider/cloud.go
index abaf8aa..beab094 100644
--- a/pkg/provider/cloud.go
+++ b/pkg/provider/cloud.go
@@ -4,16 +4,17 @@ import (
        "sigs.k8s.io/cloud-provider-kind/pkg/constants"
        "sigs.k8s.io/cloud-provider-kind/pkg/loadbalancer"
 
+       "k8s.io/apimachinery/pkg/version"
        cloudprovider "k8s.io/cloud-provider"
 
        "sigs.k8s.io/kind/pkg/cluster"
 )
 
-func New(clusterName string, kindClient *cluster.Provider) cloudprovider.Interface {
+func New(clusterName string, kindClient *cluster.Provider, ver *version.Info) cloudprovider.Interface {
        return &cloud{
                clusterName:  clusterName,
                kindClient:   kindClient,
-               lbController: loadbalancer.NewServer(),
+               lbController: loadbalancer.NewServer(ver),
        }
 }
```

There is a `CompareKubeAwareVersionStrings()`  helper on that package ...

aojea:
> > my quick test failed with non existing fields
> 
> `kubectl replace` fails with non-existing fields because it defaults to `--validate=strict`, but that's an additional behavior kubectl does (pulling the openapi and validating against it). Client-go doesn't do that.



> > my quick test failed with non existing fields
> 
> `kubectl replace` fails with non-existing fields because it defaults to `--validate=strict`, but that's an additional behavior kubectl does (pulling the openapi and validating against it). Client-go doesn't do that.

Thanks, I see this patch works against a 1.29 cluster too, so let's merge
/lgtm
/approve

PR #64, opened 2024-05-04
[WIP] KEP-1860 Make Kubernetes aware of the LoadBalancer behaviour

aojea:
Spotted by @danwinship in https://github.com/kubernetes/kubernetes/pull/124660#issuecomment-2094268703

Since cloud-provider-kind implement a loadbalancer in proxy mode, it should add the respective field.

The problem is that this field was beta in 1.30, so it will break on old clusters ... we must add a check like we did in the past for endpointslices, to obtain the version of the kubernetes cluster and add this field only for the supported versions.

This will also help to the graduation of the KEP, as it will be tested in the CI https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/1860-kube-proxy-IP-node-binding

### DIFF
--- a/pkg/loadbalancer/server.go
+++ b/pkg/loadbalancer/server.go
@@ -14,6 +14,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	cloudprovider "k8s.io/cloud-provider"
 	"k8s.io/klog/v2"
+	"k8s.io/utils/ptr"
 	"sigs.k8s.io/cloud-provider-kind/pkg/constants"
 	"sigs.k8s.io/cloud-provider-kind/pkg/container"
 )
@@ -65,10 +66,17 @@ func (s *Server) GetLoadBalancer(ctx context.Context, clusterName string, servic
 		}
 	}
 	if ipv4 != "" && svcIPv4 {
-		status.Ingress = append(status.Ingress, v1.LoadBalancerIngress{IP: ipv4, Ports: portStatus})
+		status.Ingress = append(status.Ingress, v1.LoadBalancerIngress{
+			IP:     ipv4,
+			Ports:  portStatus,
+			IPMode: ptr.To(v1.LoadBalancerIPModeProxy),
+		})
 	}
 	if ipv6 != "" && svcIPv6 {
-		status.Ingress = append(status.Ingress, v1.LoadBalancerIngress{IP: ipv6, Ports: portStatus})
+		status.Ingress = append(status.Ingress, v1.LoadBalancerIngress{
+			IP:     ipv6,
+			Ports:  portStatus,
+			IPMode: ptr.To(v1.LoadBalancerIPModeProxy)})
 	}
 
 	return status, true, nil


aojea:
/assign @danwinship @rikatz 
/hold
/help

it is missing the part that detects the kubernetes version, if anyone wants to take over feel free to duplicate this PR

pkg/loadbalancer/server.go
```diff
@@ -65,10 +66,17 @@ func (s *Server) GetLoadBalancer(ctx context.Context, clusterName string, servic
 		}
 	}
 	if ipv4 != "" && svcIPv4 {
```
aojea, line 68: This should do something like this, get the kubernetes version of the cluster (not call it each time)

```
sv, _ := kubeClient.ServerVersion()
```

and then add or skip the IPMode if is not supported as we do in kind

https://github.com/kubernetes-sigs/kind/blob/19df3db3ac3d717856f2e93537aa42b09a903706/pkg/cluster/internal/kubeadm/config.go#L477-L482

danwinship:
dup of #63 :slightly_smiling_face:

pkg/loadbalancer/server.go
danwinship: Unless we're actually going to error out if `IPMode` is not supported (which would technically be correct since the only configuration we support is `Proxy` mode with `IP` set, which was not an allowed combination pre-`IPMode`), there is no real reason to care whether the apiserver or the service proxy knows about the existence of the `IPMode` field; the result of not setting the field is exactly the same as the result of setting it and having it be ignored.

aojea: Will be it ignored?
If the API server receives an object with unknown field does it discard them?

I always forget about this, can we validate this assumption?

aojea:
> dup of #63 🙂

/close

I need to go.over my inbox,. didn't realize about it, sorry

pkg/loadbalancer/server.go
aojea: if I try to use `kubectl replace --subresource status service ..` with a field that does not exist I got this error

```
I0505 09:44:01.157687 3769194 request.go:1212] Response Body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Service in version \"v1\" cannot be handled as a Service: strict decoding error: unknown field \"status.loadBalancer.ingress[0].nonvalid\"","reason":"BadRequest","code":400}
I0505 09:44:01.159088 3769194 helpers.go:246] server response object: [{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {},
  "status": "Failure",
  "message": "error when replacing \"apache.yaml\": Service in version \"v1\" cannot be handled as a Service: strict decoding error: unknown field \"status.loadBalancer.ingress[0].nonvalid\"",
  "reason": "BadRequest",
  "code": 400
}]
Error from server (BadRequest): error when replacing "apache.yaml": Service in version "v1" cannot be handled as a Service: strict decoding error: unknown field "status.loadBalancer.ingress[0].nonvalid"

```

PR #68, opened 2024-05-16
solve UDP crashes

aojea:
This restarts the loadbalancer container when it crashes because of https://github.com/envoyproxy/envoy/issues/34195

Apparently on the restart it works fine and the test goes through.

There are a lot of hacks to deal with files and the user id for the container command is different than the one to update them, so during the restart it fails to touch the config files, since we are not worried by security on these envoys I just give them a+rw  .... the goal is to move to dynamic configuration using GRPC https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/operations/dynamic_configuration so this hack will serve until then


aojea:
ok, we are down to 2 failures now

Issue #81, opened 2024-05-30
Improve observability for loadbalancer containers


aojea:
> But that seems like it should give a pretty loud error

The envoy logs don't appear to end up in the e2e artifacts so I guess it wouldn't matter how loud it was...

_Originally posted by @danwinship in https://github.com/kubernetes-sigs/cloud-provider-kind/issues/80#issuecomment-2140982634_
            

It also could help to detect the crashes https://github.com/envoyproxy/envoy/issues/34195

I think we can add a flag for "loadbalancer logging" and indicate a file prefix to store the container logs and events, we hook into the delete operation https://github.com/kubernetes-sigs/cloud-provider-kind/blob/ae8b178e60f93b98e0d5001a47b01f003347cb8f/pkg/loadbalancer/server.go#L146 to do the `docker logs ... > file-x.log` and `docker events --filter container=CONTAINER   `

PR #82, opened 2024-05-31
log container loadbalancers

aojea:
Fixes: https://github.com/kubernetes-sigs/cloud-provider-kind/issues/81


BenTheElder (commented):
can we dump individual log files to a configurable directory on exit instead?

and also make it optional, if you're not kubernetes CI you don't need this because you can check `docker logs` interactively, but you should have the option to `--enable-log-dumping` `--logs-dir` and have the directory default to $ARTIFACTS/cloud-provider-kind/

PR #98, opened 2024-06-25
update go modules

aojea:


aojea:
> Kubernetes e2e suite: [It] [sig-api-machinery] Servers with support for API chunking should return chunks of results for list calls [Conformance] expand_more	

why is this failing?

aojea:
opened https://github.com/kubernetes/kubernetes/issues/125700

PR #99, opened 2024-06-28
adding ability to output loadbalancer status changes

calebwashburn:
Wanting to be able to watch for updates to loadbalancers so added flag to send output to system out in json format to be able to parse.  

Interested to see if there is a better way to do this where process running this binary can react to status changes for loadbalancers added/removed.

### DIFF
--- a/cmd/app.go
+++ b/cmd/app.go
@@ -18,15 +18,17 @@ import (
 )
 
 var (
-	flagV         int
-	enableLogDump bool
-	logDumpDir    string
+	flagV                    int
+	enableLogDump            bool
+	logDumpDir               string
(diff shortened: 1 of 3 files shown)


BenTheElder:
> Interested to see if there is a better way to do this where process running this binary can react to status changes for loadbalancers added/removed.

Maybe I've misread this, but if you're saying enabling another process to watch this one and react to things, that is probably not something we should enable and instead whatever reactions exist should be written as another controller with its own reconciler..?

calebwashburn:
@BenTheElder makes sense was looking for a quick way to do this but likely cleaner to a proper controller that watches for state changes.   Will close the pr.

Issue #100, opened 2024-07-02
WSL2 support


sky92zwq:
env: wsl2, ubuntu22.04
```
% go version                 
go version go1.21.4 linux/amd64
% kind version               
kind v0.20.0 go1.21.4 linux/amd64
```

```
% kubectl get svc
NAME                            TYPE           CLUSTER-IP      EXTERNAL-IP   PORT(S)               AGE
envoy-default-eg-e41e7b31       LoadBalancer   10.96.164.54    172.18.0.3    80:31407/TCP          5d20h
envoy-gateway                   ClusterIP      10.96.240.27    <none>        18000/TCP,18001/TCP   4d23h
envoy-gateway-metrics-service   ClusterIP      10.96.163.151   <none>        19001/TCP             4d23h
```
I can see the external-ip, but it will be hang when curl the ip.
`% curl 172.18.0.3:80`
And I deploy the example from [this](https://kind.sigs.k8s.io/docs/user/loadbalancer/). It is same.

aojea:
It seems this require some magic with docker https://github.com/microsoft/WSL/issues/4983#issuecomment-1560668422

What version of cloud provider kind are you using?

sky92zwq:
> It seems this require some magic with docker [microsoft/WSL#4983 (comment)](https://github.com/microsoft/WSL/issues/4983#issuecomment-1560668422)
> 
> What version of cloud provider kind are you using?

I use `go install sigs.k8s.io/cloud-provider-kind@latest`, maybe v0.3.0,(I don't find the command of cloud-provider-kind about version)
And I tried the magic, but not success until now.
Because my `ip add` shows like:
`inet 172.21.225.217/20 brd 172.21.239.255 scope global eth0`
the address is not same with external-ip above.

aojea:
should be fixed by https://github.com/kubernetes-sigs/cloud-provider-kind/pull/101

sky92zwq:
it works all fine except this [issue](https://github.com/kubernetes-sigs/cloud-provider-kind/issues/91)

yukha-dw:
Hello, does anyone know how to forward external IP from WSL2 to Windows? I can curl the IP inside WSL2, but because it is on separate network, I can't access it from Windows. Weirdly, I believe I was able to access them right away in the past

DacDinh147:
I got the same problem. I got the external IP but it hang forever when curl the ip.

```
[...]

PR #101, opened 2024-07-16
add WSL2/linux support

adone:
It's an attempt to support WSL 2

It should be run with sudo like on any other platform tunnels required:
```
sudo bin/cloud-provider-kind
```

#100

### DIFF
--- a/pkg/loadbalancer/address_other.go
+++ b/pkg/loadbalancer/address_other.go
@@ -2,10 +2,20 @@
 
 package loadbalancer
 
-func AddIPToInterface(ifaceName string, ip string) error {
+import "os/exec"
+
+func AddIPToLocalInterface(ip string) error {
+	err := exec.Command("ip", "addr", "add", ip, "dev", "lo").Run()
+	if err != nil {
+		return err
+	}
 	return nil
 }
 
-func RemoveIPToInterface(ifaceName string, ip string) error {
+func RemoveIPFromLocalInterface(ip string) error {
+	err := exec.Command("ip", "addr", "del", ip, "dev", "lo").Run()
+	if err != nil {
+		return err
+	}
 	return nil
 }
--- a/pkg/loadbalancer/address_darwin.go
+++ b/pkg/loadbalancer/address_darwin.go
@@ -6,18 +6,18 @@ import (
 	"os/exec"
 )
 
-func AddIPToInterface(ifaceName string, ip string) error {
+func AddIPToLocalInterface(ip string) error {
 	// TODO: IPv6
-	err := exec.Command("ifconfig", ifaceName, "alias", ip, "netmask", "255.255.255.255").Run()
+	err := exec.Command("ifconfig", "lo0", "alias", ip, "netmask", "255.255.255.255").Run()
 	if err != nil {
 		return err
 	}
 	return nil
 }
 
-func RemoveIPToInterface(ifaceName string, ip string) error {
+func RemoveIPFromLocalInterface(ip string) error {
 	// delete the IP address
-	err := exec.Command("ifconfig", ifaceName, "-alias", ip).Run()
+	err := exec.Command("ifconfig", "lo0", "-alias", ip).Run()
 	if err != nil {
 		return err
 	}
--- a/pkg/loadbalancer/address_windows.go
+++ b/pkg/loadbalancer/address_windows.go
@@ -6,7 +6,7 @@ import (
 	"os/exec"
 )
 
-func AddIPToInterface(ifaceName string, ip string) error {
+func AddIPToLocalInterface(ip string) error {
 	err := exec.Command("netsh", "interface", "ip", "add", "address", "loopback", ip, "255.255.255.255").Run()
 	if err != nil {
 		return err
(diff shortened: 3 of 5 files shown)


BenTheElder:
can we detect WSL2 and make this automatically do the right thing when in WSL2?

adone:
it's possible
```
v, err := ioutil.ReadFile("/proc/version")
strings.Contains(string(v), "WSL")
```
I can add a function for this `isWSL` instead of the flag

aojea:
> it's possible
> 
> ```
> v, err := ioutil.ReadFile("/proc/version")
> strings.Contains(string(v), "WSL")
> ```
> 
> I can add a function for this `isWSL` instead of the flag

please do that

aojea:
/ok-to-test

also , please confirm it works

pkg/loadbalancer/address_other.go
```diff
@@ -2,10 +2,20 @@
 
 package loadbalancer
 
+import "os/exec"
+
 func AddIPToInterface(ifaceName string, ip string) error {
+	err := exec.Command("ip", "addr", "add", ip, "dev", "lo").Run()
```
aojea, line 8:
```suggestion
	err := exec.Command("ip", "addr", "add", ip, "dev", ifaceName).Run()
```

pkg/loadbalancer/address_other.go
```diff
@@ -2,10 +2,20 @@
...
+	if err != nil {
+		return err
+	}
 	return nil
 }
 
 func RemoveIPToInterface(ifaceName string, ip string) error {
+	err := exec.Command("ip", "addr", "del", ip, "dev", "lo").Run()
```
aojea, line 16:
```suggestion
	err := exec.Command("ip", "addr", "del", ip, "dev", ifaceName).Run()
```

pkg/loadbalancer/address_other.go
adone: unfortunately it doesn't work:
```
failed to ensure load balancer: exit status 1
```
because in WSL2 there is only `lo`
```
❯ ifconfig
eth0: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 1500
        inet 172.25.124.246  netmask 255.255.240.0  broadcast 172.25.127.255
        inet6 fe80::215:5dff:feff:b7a0  prefixlen 64  scopeid 0x20<link>
        ether 00:15:5d:ff:b7:a0  txqueuelen 1000  (Ethernet)
        RX packets 76102  bytes 87555200 (87.5 MB)
        RX errors 0  dropped 0  overruns 0  frame 0
        TX packets 59909  bytes 7699849 (7.6 MB)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0

lo: flags=73<UP,LOOPBACK,RUNNING>  mtu 65536
        inet 127.0.0.1  netmask 255.0.0.0
        inet6 ::1  prefixlen 128  scopeid 0x10<host>
        loop  txqueuelen 1000  (Local Loopback)
        RX packets 646306  bytes 1683348381 (1.6 GB)
        RX errors 0  dropped 0  overruns 0  frame 0
        TX packets 646306  bytes 1683348381 (1.6 GB)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0
```
and with `ifaceName` == `lo0`
```
❯ sudo ip addr add 172.20.0.3 dev lo0
Cannot find device "lo0"
```

aojea: then do something like this
```diff
diff --git a/pkg/loadbalancer/tunnel.go b/pkg/loadbalancer/tunnel.go
index 058e106..3fd4b62 100644
--- a/pkg/loadbalancer/tunnel.go
+++ b/pkg/loadbalancer/tunnel.go
@@ -43,6 +43,11 @@ func (t *tunnelManager) setupTunnels(containerName string) error {
        }
 
        klog.V(0).Infof("setting IPv4 address %s associated to container %s", ipv4, containerName)
+       // linux and windows uses lo0
+       ifaceName := "lo0"
+       if isWSL2() {
+               ifaceName = "lo"
+       }
        err = AddIPToInterface(ifaceName, ipv4)
        if err != nil {
                return err
```

but respect the API

aojea:
almost there, left a suggestion and then  please squash all commits

pkg/loadbalancer/address_other.go
adone: `lo0` is a name of a local interface in MacOS only
even in `address_windows.go` it isn't used (it's replaced with hardcoded `loopback`):
```
err := exec.Command("netsh", "interface", "ip", "add", "address", "loopback", ip, "255.255.255.255").Run()
```
so why it is a default value for the constant? 

I suggest to move `ifaceName` to implementation specific layer (`address_*.go` files) and remove parameter from functions `AddIPToInterface` and `RemoveIPToInterface`
they can also be renamed to something like `AddIPToLocalInterface` and `RemoveIPFromLocalInterface` accordingly

adone: done

aojea:
/lgtm
/approve

Thanks for your patience and for improving the project

BenTheElder:
thanks!

adone:
late reply but anyway
yep, can access services via localhost and random port from windows/browser and via curl to EXTERNAL IP from WSL2

Issue #111, opened 2024-08-03
envoy failed to start on linux servers without ipv6


rophy:
- host: Ubuntu 22.04.3 LTS
- docker: 24.0.7
- cloud-provider-kind: 0.3.0

```console
E0803 14:16:18.014910       1 controller.go:298] error processing service istio-gateway/istio-gateway (retrying with exponential backoff): failed to ensure load balancer: failed to create continers kindccm-MDL2VIYWAK3ZKKYF6N7NN3NLC7W5RZ4JL24CY7PK [--detach --tty --label io.x-k8s.cloud-provider-kind.cluster=kind --label io.x-k8s.cloud-provider-kind.loadbalancer.name=kind/istio-gateway/istio-gateway --net kind --init=false --hostname kindccm-MDL2VIYWAK3ZKKYF6N7NN3NLC7W5RZ4JL24CY7PK --privileged --restart=on-failure --sysctl=net.ipv4.ip_forward=1 --sysctl=net.ipv6.conf.all.disable_ipv6=0 --sysctl=net.ipv6.conf.all.forwarding=1 --sysctl=net.ipv4.conf.all.rp_filter=0 --publish=10000/TCP --publish-all envoyproxy/envoy:v1.30.1 bash -c echo -en 'node:
  cluster: cloud-provider-kind
  id: cloud-provider-kind-id

dynamic_resources:
  cds_config:
    resource_api_version: V3
    path: /home/envoy/cds.yaml
  lds_config:
    resource_api_version: V3
    path: /home/envoy/lds.yaml


admin:
  access_log_path: /dev/stdout
  address:
    socket_address:
      address: 0.0.0.0
      port_value: 10000
' > /home/envoy/envoy.yaml && touch /home/envoy/cds.yaml && touch /home/envoy/lds.yaml && while true; do envoy -c /home/envoy/envoy.yaml && break; sleep 1; done]: exit status 127
```

Trying to run the docker run command as shown in the log, I can see:

```console
[...]

PR #114, opened 2024-08-06
Only enable IPv6 on loadbalancer container when the service is ipv6

aojea:
Fixes: https://github.com/kubernetes-sigs/cloud-provider-kind/issues/111

### DIFF
--- a/pkg/loadbalancer/proxy.go
+++ b/pkg/loadbalancer/proxy.go
@@ -329,7 +329,7 @@ func waitLoadBalancerReady(ctx context.Context, name string, timeout time.Durati
 		// iptables port forwarding on localhost only works for IPv4
 		resp, err := httpClient.Get(fmt.Sprintf("http://127.0.0.1:%s/ready", port))
 		if err != nil {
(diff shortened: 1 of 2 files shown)


aojea:
> Kubernetes e2e suite: [It] [sig-network] LoadBalancers ExternalTrafficPolicy: Local [Feature:LoadBalancer] [Slow] should handle updates to ExternalTrafficPolicy field expand_less	

the test failure is a known issue, unrelated to this PR

BenTheElder (commented):
/lgtm
/approve

BenTheElder:
hmm @aojea probably some more of these presubmits should be marked required?

aojea:
> hmm @aojea probably some more of these presubmits should be marked required?

Indeed

serbaut:
I think this issue has to be handled [here ](https://github.com/kubernetes-sigs/cloud-provider-kind/blob/main/pkg/gateway/envoy.go#L192 )as well. envoy fails to start on a host with ipv6 disabled.

Issue #116, opened 2024-08-18
failed to create containers kindccm on applying LoadBalancer example with kubectl 


LogExE:
I get error when trying to run an example from https://kind.sigs.k8s.io/docs/user/loadbalancer/
I use podman driver. My system is Fedora Silverblue 40.

Steps to reproduce:
1. Be on latest Fedora Silverblue version
2. Run `sudo kind create cluster`
3. Run `sudo cloud-provider-kind`
4. Execute `sudo kubectl apply` for manifest https://kind.sigs.k8s.io/examples/loadbalancer/usage.yaml
5. Now cloud-provider-kind constantly prints error on failing to create container

The error:
```sh
E0818 16:24:32.239123   25489 controller.go:298] error processing service default/foo-service (retrying with exponential backoff): failed to ensure load balancer: failed to create continers kindccm-TLRDKPBWWH4DUSI7J7BNE3ABETEPCKSYA6UIWR5B [--detach --tty --label io.x-k8s.cloud-provider-kind.cluster=kind --label io.x-k8s.cloud-provider-kind.loadbalancer.name=kind/default/foo-service --net kind --init=false --hostname kindccm-TLRDKPBWWH4DUSI7J7BNE3ABETEPCKSYA6UIWR5B --privileged --restart=on-failure --sysctl=net.ipv4.ip_forward=1 --sysctl=net.ipv6.conf.all.disable_ipv6=0 --sysctl=net.ipv6.conf.all.forwarding=1 --sysctl=net.ipv4.conf.all.rp_filter=0 --publish=10000/TCP --publish-all envoyproxy/envoy:v1.30.1 bash -c echo -en 'node:
  cluster: cloud-provider-kind
  id: cloud-provider-kind-id

dynamic_resources:
  cds_config:
    resource_api_version: V3
    path: /home/envoy/cds.yaml
  lds_config:
    resource_api_version: V3
    path: /home/envoy/lds.yaml

admin:
  access_log_path: /dev/stdout
  address:
    socket_address:
      address: 0.0.0.0
      port_value: 10000
' > /home/envoy/envoy.yaml && touch /home/envoy/cds.yaml && touch /home/envoy/lds.yaml && while true; do envoy -c /home/envoy/envoy.yaml && break; sleep 1; done]: exit status 125
```

aojea:
Interesting ... can you list what containers are actually running ?
I think it is something as `podman ps -a` and the networks `podman network ls`?

[...]

PR #124, opened 2024-08-20
Use fully qualified image name for envoy proxy image

vially:
This should fix an issue where `docker.io` might not be in the list of `unqualified-search-registries` when using Podman.

This _might_ be the root cause for #116.

### DIFF
--- a/pkg/loadbalancer/proxy.go
+++ b/pkg/loadbalancer/proxy.go
@@ -21,7 +21,7 @@ import (
 )
 
 // proxyImage defines the loadbalancer image:tag
-const proxyImage = "envoyproxy/envoy:v1.30.1"
+const proxyImage = "docker.io/envoyproxy/envoy:v1.30.1"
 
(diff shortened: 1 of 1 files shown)


BenTheElder:
Aside: you could coin a term "fully qualified image", but we are using the term FQDN incorrectly, would imply `docker.io.` (and therefore no DNS search), which we're not doing here (and shouldn't, I don't think it's valid in image names and would probably break things)

BenTheElder (commented):
/ok-to-test
/lgtm

thanks!

vially:
> Aside: you could coin a term "fully qualified image", but we are using the term FQDN incorrectly, would imply `docker.io.` (and therefore no DNS search), which we're not doing here (and shouldn't, I don't think it's valid in image names and would probably break things)

Good point. I've updated the title and replaced FQDN with "fully qualified image name" which seems more accurate.

Issue #109, opened 2024-08-03
Running cloud-provider-kind 0.3.0 as container fails to probe ccm healthcheck


rophy:
## Env Info

```console
default:~/containers/cloud-provider-kind$ kind version
kind v0.14.0 go1.18.2 linux/amd64

default:~/containers/cloud-provider-kind$ docker version
Client: Docker Engine - Community
 Version:           27.0.3
 API version:       1.46
 Go version:        go1.21.11
 Git commit:        7d4bcd8
 Built:             Sat Jun 29 00:02:33 2024
 OS/Arch:           linux/amd64
 Context:           default

Server: Docker Engine - Community
 Engine:
  Version:          27.0.3
  API version:      1.46 (minimum version 1.24)
  Go version:       go1.21.11
  Git commit:       662f78c
  Built:            Sat Jun 29 00:02:33 2024
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.7.19
  GitCommit:        2bf793ef6dc9a18e00cb12efb64355c2c9d5eb41
 runc:
  Version:          1.7.19
  GitCommit:        v1.1.13-0-g58aa920
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0.
```

## Steps to reproduce

1. `kind create cluster`
2. `docker run --rm --network kind -v /var/run/docker.sock:/var/run/docker.sock cloud-provider-kind`
3. Create a LoadBalancer service following [README](https://github.com/kubernetes-sigs/cloud-provider-kind?tab=readme-ov-file#creating-a-service-and-exposing-it-via-a-loadbalancer)

## Expected Result

cloud-provider-kind should be able to:

1. Detect the creation of a LB service
2. Create a ccm container
3. Probe its readiness as success

## Actual Result

cloud-provider-kind can do step 1, 2, and fails on step 3

```console
I0803 02:16:57.571892       1 instances.go:47] Check instance metadata for kind-control-plane
[...]

PR #125, opened 2024-08-22
Fix load balancer readiness checks when running inside a container.

Dadeos-Menlo:
I encountered the issues reported in #109. However, because I am running the system on macOS, using the `--network host` option when launching the "cloud-provider-kind" container doesn't work. Use of the "kind" defined network is required in order for the "cloud-provider-kind" container to successfully [communicate](https://github.com/kubernetes-sigs/cloud-provider-kind/blob/a384948631c160bf5c9732c01bff0641c224f400/pkg/controller/controller.go#L123) with the Kind Kubernetes controller (i.e. `internal` must be `true` - I suspect that the [comment](https://github.com/kubernetes-sigs/cloud-provider-kind/blob/a384948631c160bf5c9732c01bff0641c224f400/pkg/controller/controller.go#L121) is incorrect, `internal == false` is tried first…):
```
11:24:12.042101       1 app.go:41] FLAG: --enable-log-dumping="false"
11:24:12.042161       1 app.go:41] FLAG: --logs-dir=""
11:24:12.042168       1 app.go:41] FLAG: --v="2"
11:24:12.266173       1 controller.go:166] probe HTTP address https://127.0.0.1:51518
11:24:12.266644       1 controller.go:169] Failed to connect to HTTP address https://127.0.0.1:51518: Get "https://127.0.0.1:51518": dial tcp 127.0.0.1:51518: connect: connection refused
11:24:12.266711       1 controller.go:166] probe HTTP address https://127.0.0.1:51518
11:24:12.267021       1 controller.go:169] Failed to connect to HTTP address https://127.0.0.1:51518: Get "https://127.0.0.1:51518": dial tcp 127.0.0.1:51518: connect: connection refused
11:24:13.270465       1 controller.go:166] probe HTTP address https://127.0.0.1:51518
11:24:13.271440       1 controller.go:169] Failed to connect to HTTP address https://127.0.0.1:51518: Get "https://127.0.0.1:51518": dial tcp 127.0.0.1:51518: connect: connection refused
11:24:15.272618       1 controller.go:166] probe HTTP address https://127.0.0.1:51518
11:24:15.273070       1 controller.go:169] Failed to connect to HTTP address https://127.0.0.1:51518: Get "https://127.0.0.1:51518": dial tcp 127.0.0.1:51518: connect: connection refused
11:24:18.275251       1 controller.go:166] probe HTTP address https://127.0.0.1:51518
11:24:18.275680       1 controller.go:169] Failed to connect to HTTP address https://127.0.0.1:51518: Get "https://127.0.0.1:51518": dial tcp 127.0.0.1:51518: connect: connection refused
11:24:22.276154       1 controller.go:151] Failed to connect to apiserver kind: <nil>
11:24:22.370267       1 controller.go:166] probe HTTP address https://kind-control-plane:6443
11:24:22.376355       1 controller.go:83] Creating new cloud provider for cluster kind
11:24:22.382101       1 controller.go:90] Starting cloud controller for cluster kind
```

The proposed changes seek to identify if/when the system is executing within a container. There doesn't appear to be a great universal way of achieving this, so I went with defining a `container` environment variable within the Dockerfile, inspired by [systemd's Container Interface specification](https://systemd.io/CONTAINER_INTERFACE/#environment-variables). An alternative approach might be to pass a command-line [flag](https://pkg.go.dev/flag), but plumbing such a value through to where it's required seemed non-trivial. I also found it necessary to enable the [tunnelManager](https://github.com/kubernetes-sigs/cloud-provider-kind/blob/a384948631c160bf5c9732c01bff0641c224f400/pkg/loadbalancer/server.go#L32L34) implementation for the case where the system is running on Linux within a container.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -74,14 +74,14 @@ func (c *Controller) Run(ctx context.Context) {
 				continue
 			}
 
-			kubeClient, err := c.getKubeClient(ctx, cluster)
+			kubeClient, directConnectivity, err := c.getKubeClient(ctx, cluster)
 			if err != nil {
 				klog.Errorf("Failed to create kubeClient for cluster %s: %v", cluster, err)
 				continue
 			}
 
 			klog.V(2).Infof("Creating new cloud provider for cluster %s", cluster)
-			cloud := provider.New(cluster, c.kind)
+			cloud := provider.New(cluster, c.kind, directConnectivity)
 			ccm, err := startCloudControllerManager(ctx, cluster, kubeClient, cloud)
 			if err != nil {
 				klog.Errorf("Failed to start cloud controller for cluster %s: %v", cluster, err)
@@ -111,14 +111,13 @@ func (c *Controller) Run(ctx context.Context) {
 // getKubeClient returns a kubeclient depending if the ccm runs inside a container
 // inside the same docker network that the kind cluster or run externally in the host
 // It tries first to connect to the external endpoint
-func (c *Controller) getKubeClient(ctx context.Context, cluster string) (kubernetes.Interface, error) {
+func (c *Controller) getKubeClient(ctx context.Context, cluster string) (kubernetes.Interface, bool, error) {
 	httpClient := &http.Client{
 		Timeout: 5 * time.Second,
 		Transport: &http.Transport{
 			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
 		},
 	}
-	// try internal first
 	for _, internal := range []bool{false, true} {
 		kconfig, err := c.kind.KubeConfig(cluster, internal)
 		if err != nil {
@@ -138,7 +137,7 @@ func (c *Controller) getKubeClient(ctx context.Context, cluster string) (kuberne
 		for i := 0; i < 5; i++ {
 			select {
 			case <-ctx.Done():
-				return nil, ctx.Err()
+				return nil, false, ctx.Err()
 			default:
 			}
 			if probeHTTP(httpClient, config.Host) {
@@ -157,9 +156,9 @@ func (c *Controller) getKubeClient(ctx context.Context, cluster string) (kuberne
 			klog.Errorf("Failed to create kubeClient for cluster %s: %v", cluster, err)
 			continue
 		}
-		return kubeClient, err
+		return kubeClient, internal, err
 	}
-	return nil, fmt.Errorf("can not find a working kubernetes clientset")
+	return nil, false, fmt.Errorf("can not find a working kubernetes clientset")
 }
 
 func probeHTTP(client *http.Client, address string) bool {
--- a/pkg/loadbalancer/proxy.go
+++ b/pkg/loadbalancer/proxy.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"strconv"
 	"strings"
@@ -274,7 +275,7 @@ func generateConfig(service *v1.Service, nodes []*v1.Node) *proxyConfigData {
 }
 
 // TODO: move to xDS via GRPC instead of having to deal with files
-func proxyUpdateLoadBalancer(ctx context.Context, clusterName string, service *v1.Service, nodes []*v1.Node) error {
+func proxyUpdateLoadBalancer(ctx context.Context, clusterName string, service *v1.Service, nodes []*v1.Node, directConnectivity bool) error {
 	if service == nil {
 		return nil
 	}
@@ -311,23 +312,34 @@ func proxyUpdateLoadBalancer(ctx context.Context, clusterName string, service *v
 	if err != nil {
 		return fmt.Errorf("error updating configuration Stdout: %s Stderr: %s : %w", stdout.String(), stderr.String(), err)
 	}
-	return waitLoadBalancerReady(ctx, name, 30*time.Second)
+	return waitLoadBalancerReady(ctx, name, 30*time.Second, directConnectivity)
 }
 
-func waitLoadBalancerReady(ctx context.Context, name string, timeout time.Duration) error {
+func waitLoadBalancerReady(ctx context.Context, name string, timeout time.Duration, directConnectivity bool) error {
 	portmaps, err := container.PortMaps(name)
 	if err != nil {
 		return err
 	}
-	port, ok := portmaps[strconv.Itoa(envoyAdminPort)]
-	if !ok {
-		return fmt.Errorf("envoy admin port %d not found, got %v", envoyAdminPort, portmaps)
+
+	var authority string
+	if directConnectivity {
+		ipv4, _, err := container.IPs(name)
+		if err != nil {
+			return err
+		}
+		authority = net.JoinHostPort(ipv4, strconv.Itoa(envoyAdminPort))
+	} else {
+		port, ok := portmaps[strconv.Itoa(envoyAdminPort)]
+		if !ok {
+			return fmt.Errorf("envoy admin port %d not found, got %v", envoyAdminPort, portmaps)
+		}
+		authority = net.JoinHostPort("127.0.0.1", port)
 	}
 
 	httpClient := http.DefaultClient
 	err = wait.PollUntilContextTimeout(ctx, 1*time.Second, timeout, true, func(ctx context.Context) (done bool, err error) {
 		// iptables port forwarding on localhost only works for IPv4
-		resp, err := httpClient.Get(fmt.Sprintf("http://127.0.0.1:%s/ready", port))
+		resp, err := httpClient.Get(fmt.Sprintf("http://%s/ready", authority))
 		if err != nil {
 			klog.V(2).Infof("unexpected error trying to get load balancer %s readiness :%v", name, err)
 			return false, nil
--- a/pkg/loadbalancer/server.go
+++ b/pkg/loadbalancer/server.go
@@ -22,14 +22,17 @@ import (
 )
 
 type Server struct {
+	directConnectivity bool
 	tunnelManager *tunnelManager
 }
 
 var _ cloudprovider.LoadBalancer = &Server{}
 
-func NewServer() cloudprovider.LoadBalancer {
-	s := &Server{}
-	if runtime.GOOS == "darwin" || runtime.GOOS == "windows" || isWSL2() {
+func NewServer(directConnectivity bool) cloudprovider.LoadBalancer {
+	s := &Server{
+		directConnectivity: directConnectivity,
+	}
+	if directConnectivity || runtime.GOOS == "darwin" || runtime.GOOS == "windows" || isWSL2() {
 		s.tunnelManager = NewTunnelManager()
 	}
 	return s
@@ -145,7 +148,7 @@ func (s *Server) EnsureLoadBalancer(ctx context.Context, clusterName string, ser
 }
 
 func (s *Server) UpdateLoadBalancer(ctx context.Context, clusterName string, service *v1.Service, nodes []*v1.Node) error {
-	return proxyUpdateLoadBalancer(ctx, clusterName, service, nodes)
+	return proxyUpdateLoadBalancer(ctx, clusterName, service, nodes, s.directConnectivity)
 }
 
 func (s *Server) EnsureLoadBalancerDeleted(ctx context.Context, clusterName string, service *v1.Service) error {
--- a/pkg/provider/cloud.go
+++ b/pkg/provider/cloud.go
@@ -9,11 +9,11 @@ import (
 	"sigs.k8s.io/kind/pkg/cluster"
 )
 
-func New(clusterName string, kindClient *cluster.Provider) cloudprovider.Interface {
+func New(clusterName string, kindClient *cluster.Provider, directConnectivity bool) cloudprovider.Interface {
 	return &cloud{
 		clusterName:  clusterName,
 		kindClient:   kindClient,
-		lbController: loadbalancer.NewServer(),
+		lbController: loadbalancer.NewServer(directConnectivity),
 	}
 }
 


Dadeos-Menlo:
> if you run in mac, and cloud provider kind is running in a container inside the VM ... are you still be able to connect to the exposed ports from the loadbalancer?

Yes, via the [tunnelManager](https://github.com/kubernetes-sigs/cloud-provider-kind/blob/a384948631c160bf5c9732c01bff0641c224f400/pkg/loadbalancer/server.go#L32L34) implementation…

For example; I've defined a Cilium [Gateway](https://docs.cilium.io/en/stable/network/servicemesh/gateway-api/gateway-api/#gateway-api-support) bound to port `14000`, which I am able to access from the host (i.e. my Mac) via the port mapped by Docker (i.e. `0.0.0.0:32774->14000/tcp`): 
```
% curl http://localhost:32774/
```
```
% docker ps
CONTAINER ID   IMAGE                      COMMAND                  CREATED              STATUS              PORTS                                                                                          NAMES
9af2817a1f63   envoyproxy/envoy:v1.30.1   "/docker-entrypoint.…"   22 seconds ago       Up 22 seconds       0.0.0.0:32773->10000/tcp, :::32773->10000/tcp, 0.0.0.0:32774->14000/tcp, :::32774->14000/tcp   kindccm-S4I65OVHZT6XWEPYZ7W72XP3U3LZGKGKFNJHMLZH
ca8be8fc03c5   cloud-provider-kind        "/bin/cloud-provider…"   About a minute ago   Up About a minute   2375-2376/tcp                                                                                  cloud-provider-kind
```
I presume that this is the intended behaviour?

I wish to run "cloud-provider-kind" in a container so that I can use:
```
% docker container run --detach --restart=always --network 'kind' --volume '/var/run/docker.sock:/var/run/docker.sock' --name 'cloud-provider-kind' 'cloud-provider-kind'
```
and largely forget about it (i.e. have the "cloud-provider-kind" container automagically restart, just like the Kind node container[s] do, whenever I restart my Mac and/or the Linux VM running the Docker daemon); so all I need worry about is working out the port mapping that Docker has created for the Envoy proxy container backing the load balancer.

aojea:
Ok, it took me a while to parse this sorry, yeah, I think the comment `internal first` is wrong.

I think there are two variables here:

- platforms that run containers in a VM needs tunnels to get to the containers from the host: Mac, Windows, WSL2
- where cloud-provider-kind is actually running

The problem seems to be that on platforms that use tunnels, when cloud-provider-kind runs in a container, can not access the portmapped entries.

> The proposed changes seek to identify if/when the system is executing within a container

We know at the moment of polling the apiserver of the cluster what type of connectivity we have, internal false or true, is not really about "running in a container" is more about how we "can connect to the containers"

I think we can improve this by passing the connectivity type: direct or via portmap to the cloud provider loadbalancer controller implementation, a strawman approach will be something like

```diff
diff --git a/pkg/controller/controller.go b/pkg/controller/controller.go
index 329ae85..0bce8e8 100644
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -74,14 +74,14 @@ func (c *Controller) Run(ctx context.Context) {
                                continue
                        }
 
-                       kubeClient, err := c.getKubeClient(ctx, cluster)
+                       kubeClient, err, directConnectivity := c.getKubeClient(ctx, cluster)
                        if err != nil {
                                klog.Errorf("Failed to create kubeClient for cluster %s: %v", cluster, err)
                                continue
                        }
 
                        klog.V(2).Infof("Creating new cloud provider for cluster %s", cluster)
-                       cloud := provider.New(cluster, c.kind)
+                       cloud := provider.New(cluster, c.kind, directConnectivity)
                        ccm, err := startCloudControllerManager(ctx, cluster, kubeClient, cloud)
                        if err != nil {
                                klog.Errorf("Failed to start cloud controller for cluster %s: %v", cluster, err)
@@ -111,7 +111,7 @@ func (c *Controller) Run(ctx context.Context) {
 // getKubeClient returns a kubeclient depending if the ccm runs inside a container
 // inside the same docker network that the kind cluster or run externally in the host
 // It tries first to connect to the external endpoint
-func (c *Controller) getKubeClient(ctx context.Context, cluster string) (kubernetes.Interface, error) {
+func (c *Controller) getKubeClient(ctx context.Context, cluster string) (kubernetes.Interface, error, bool) {
        httpClient := &http.Client{
                Timeout: 5 * time.Second,
                Transport: &http.Transport{
@@ -138,7 +138,7 @@ func (c *Controller) getKubeClient(ctx context.Context, cluster string) (kuberne
                for i := 0; i < 5; i++ {
                        select {
                        case <-ctx.Done():
-                               return nil, ctx.Err()
+                               return nil, ctx.Err(), false
                        default:
                        }
                        if probeHTTP(httpClient, config.Host) {
@@ -157,9 +157,9 @@ func (c *Controller) getKubeClient(ctx context.Context, cluster string) (kuberne
                        klog.Errorf("Failed to create kubeClient for cluster %s: %v", cluster, err)
                        continue
                }
-               return kubeClient, err
+               return kubeClient, err, true
        }
-       return nil, fmt.Errorf("can not find a working kubernetes clientset")
+       return nil, fmt.Errorf("can not find a working kubernetes clientset"), false
 }
 
 func probeHTTP(client *http.Client, address string) bool {
diff --git a/pkg/loadbalancer/proxy.go b/pkg/loadbalancer/proxy.go
index e239ef4..a4d2c8e 100644
--- a/pkg/loadbalancer/proxy.go
+++ b/pkg/loadbalancer/proxy.go
@@ -274,7 +274,7 @@ func generateConfig(service *v1.Service, nodes []*v1.Node) *proxyConfigData {
 }
 
 // TODO: move to xDS via GRPC instead of having to deal with files
-func proxyUpdateLoadBalancer(ctx context.Context, clusterName string, service *v1.Service, nodes []*v1.Node) error {
+func proxyUpdateLoadBalancer(ctx context.Context, clusterName string, service *v1.Service, nodes []*v1.Node, directConnectivity bool) error {
        if service == nil {
                return nil
        }
@@ -311,10 +311,10 @@ func proxyUpdateLoadBalancer(ctx context.Context, clusterName string, service *v
        if err != nil {
                return fmt.Errorf("error updating configuration Stdout: %s Stderr: %s : %w", stdout.String(), stderr.String(), err)
        }
-       return waitLoadBalancerReady(ctx, name, 30*time.Second)
+       return waitLoadBalancerReady(ctx, name, 30*time.Second, directConnectivity)
 }
 
-func waitLoadBalancerReady(ctx context.Context, name string, timeout time.Duration) error {
+func waitLoadBalancerReady(ctx context.Context, name string, timeout time.Duration, directConnectivity bool) error {
        portmaps, err := container.PortMaps(name)
        if err != nil {
                return err
diff --git a/pkg/loadbalancer/server.go b/pkg/loadbalancer/server.go
index 3075c8f..8a2e33c 100644
--- a/pkg/loadbalancer/server.go
+++ b/pkg/loadbalancer/server.go
@@ -22,13 +22,16 @@ import (
 )
 
 type Server struct {
-       tunnelManager *tunnelManager
+       tunnelManager      *tunnelManager
+       directConnectivity bool
 }
 
 var _ cloudprovider.LoadBalancer = &Server{}
 
-func NewServer() cloudprovider.LoadBalancer {
-       s := &Server{}
+func NewServer(directConnectivity bool) cloudprovider.LoadBalancer {
+       s := &Server{
+               directConnectivity: directConnectivity,
+       }
        if runtime.GOOS == "darwin" || runtime.GOOS == "windows" || isWSL2() {
                s.tunnelManager = NewTunnelManager()
        }
@@ -145,7 +148,7 @@ func (s *Server) EnsureLoadBalancer(ctx context.Context, clusterName string, ser
 }
 
 func (s *Server) UpdateLoadBalancer(ctx context.Context, clusterName string, service *v1.Service, nodes []*v1.Node) error {
-       return proxyUpdateLoadBalancer(ctx, clusterName, service, nodes)
+       return proxyUpdateLoadBalancer(ctx, clusterName, service, nodes, s.directConnectivity)
 }
 
 func (s *Server) EnsureLoadBalancerDeleted(ctx context.Context, clusterName string, service *v1.Service) error {
diff --git a/pkg/provider/cloud.go b/pkg/provider/cloud.go
index abaf8aa..1f61dea 100644
--- a/pkg/provider/cloud.go
+++ b/pkg/provider/cloud.go
@@ -9,11 +9,13 @@ import (
        "sigs.k8s.io/kind/pkg/cluster"
 )
 
-func New(clusterName string, kindClient *cluster.Provider) cloudprovider.Interface {
+func New(clusterName string, kindClient *cluster.Provider, directConnectivity bool) cloudprovider.Interface {
        return &cloud{
-               clusterName:  clusterName,
-               kindClient:   kindClient,
-               lbController: loadbalancer.NewServer(),
+               clusterName: clusterName,
+               kindClient:  kindClient,
+               lbController: loadbalancer.NewServer(
+                       directConnectivity,
+               ),
        }
 }
```

We want to do better abstractions with a custom Config struct or functional options pattern to avoid keep modifying the methods https://golang.cafe/blog/golang-functional-options-pattern.html

Dadeos-Menlo:
I _think_ that https://github.com/kubernetes-sigs/cloud-provider-kind/commit/8a7b2c319584e177d1276940d459240c73497531 represents your [suggestion](https://github.com/kubernetes-sigs/cloud-provider-kind/pull/125#issuecomment-2304708444)…

Essentially there are a number of distinct deployment scenarios that you're seeking to support:
* Running on a Linux host
* Running on a non-Linux host (i.e. within a Linux VM, where tunnelling is implied)
* Running within a container with "host" networking
* Running within a container with non-"host" networking (e.g. `--network kind`)

https://github.com/kubernetes-sigs/cloud-provider-kind/commit/8a7b2c319584e177d1276940d459240c73497531 observes whether communication with the Kubernetes controller is established via `internal` or not:

https://github.com/kubernetes-sigs/cloud-provider-kind/blob/a384948631c160bf5c9732c01bff0641c224f400/pkg/controller/controller.go#L122-L123

the value of `internal` is then carried via the `directConnectivity` argument - this _might_ be a little ambiguous, I took  `directConnectivity == internal` but you _may_ have intended `directConnectivity != internal`?

When running on a Linux host, either natively or within a container using `--network host`, then `internal` is `false` and communication may be established via IP addresses assigned to the host. When running on a non-Linux host, either natively or within a container using `--network kind`, then `internal` is `true` and communication may be established via ports mapped by Docker.

pkg/loadbalancer/proxy.go
```diff
@@ -311,23 +311,34 @@ func proxyUpdateLoadBalancer(ctx context.Context, clusterName string, service *v
...
-	if !ok {
-		return fmt.Errorf("envoy admin port %d not found, got %v", envoyAdminPort, portmaps)
+
+	var authority string
+	if directConnectivity {
+		ipv4, _, err := container.IPs(name)
+		if err != nil {
+			return err
+		}
+		authority = fmt.Sprintf("%s:%s", ipv4, strconv.Itoa(envoyAdminPort))
```
aojea, line 329: use net.JoinHostPort() to build the authority so it works later for ipv6

Dadeos-Menlo:
> Can you also please check locally it works? I don't have a mac to test it

Yep; works on my machine…

I've tested running natively (on my Mac) and running within a container using `--network kind` and both behave as I require (i.e. Docker automagically maps the `14000` port of the Gateway through to the host (i.e. my Mac). I've also tested running within a container with `--network host`, where an IP address is assigned within the Linux VM, which I can verify by SSHing into the VM (I'm using [Lima](https://lima-vm.io/docs/reference/limactl_shell/)).

I haven't tested running natively on Linux (I haven't bothered to build the Linux variant of "cloud-provider-kind"). I _believe_ that you _could_ test the behaviour - presumably on a Linux host - by launching a container with `--network kind` and observing correct operation (i.e. the "ready" checks working on the Envoy proxy and Docker automagically mapping ports).

pkg/loadbalancer/proxy.go
Dadeos-Menlo: Done in 5e03531

aojea:
> I haven't tested running natively on Linux (I haven't bothered to build the Linux variant of "cloud-provider-kind"). I _believe_ that you _could_ test the behaviour - presumably on a Linux host - by launching a container with `--network kind` and observing correct operation (i.e. the "ready" checks working on the Envoy proxy and Docker automagically mapping ports).

we have CI for that, just the scenario you mentioned that failed was enough ... thanks for the effort anyway

/retest

You need to sign the CLA

pkg/loadbalancer/server.go
```diff
@@ -22,14 +22,17 @@ import (
...
 var _ cloudprovider.LoadBalancer = &Server{}
 
-func NewServer() cloudprovider.LoadBalancer {
-	s := &Server{}
-	if runtime.GOOS == "darwin" || runtime.GOOS == "windows" || isWSL2() {
+func NewServer(directConnectivity bool) cloudprovider.LoadBalancer {
+	s := &Server{
+		directConnectivity: directConnectivity,
+	}
+	if directConnectivity || runtime.GOOS == "darwin" || runtime.GOOS == "windows" || isWSL2() {
```
aojea, line 35: hmm, this is `directConnectivy && containers running inside VM` 

I don't know from the top of my head how to detect `containers running inside VM` 🤔

aojea: @BenTheElder you have any idea on how to identify if the containers are running inside a VM as in Mac or Windows?

Dadeos-Menlo: > … how to identify if the containers are running inside a VM as in Mac or Windows?

I'm not sure that that is a _valid_ distinction. The system either runs within a container or it doesn't; whether the container runtime is hosted directly on the _host_ or within a VM running on the _host_ is not really its concern?

When the readme states:

https://github.com/kubernetes-sigs/cloud-provider-kind/blob/f26b241ce573c53e81dfca8070167b0288626c03/README.md?plain=1#L176-L182

I took that to mean that the expectation should be that any load balancer _ports_ would be mapped, by Docker, to ports on the host; hence my proposal to enable the `tunnelManager` for the `directConnectivity ` scenario

https://github.com/kubernetes-sigs/cloud-provider-kind/blob/5e03531c9674276a444e7270a1452bfc3c03a048/pkg/loadbalancer/server.go#L35-L37

I don't understand why anyone would not wish to run the "cloud-provider-kind" system within a container - because `--restart=always` can be used, thus giving the system the same scope/lifetime as any corresponding Kind clusters. So, it strikes me that, the pertinent distinction is whether `--network host` or `--network kind` is used when running the container, which respectively correspond to the value of `internal`/`directConnectivity`. If running on a Linux _host_ then `--network host` would be used, because that facilitates direct IP and port registration behaviour, whereas a non-Linux _host_ is required to use `--network kind`, thus limiting the functionality to the host port maps created by Docker.

Dadeos-Menlo:
Signing of the CLA is in train; but I'll require the corporate variant, which is a non-trivial exercise…

pkg/loadbalancer/server.go
aojea: I'm just a pretty busy these days, I want this change to land but I want also to get this as much scoped as possible to avoid possible edge cases and to have sane defaults, relying on users to follow recommendations never works XD ... just please have some patience, I will make time next week to finish this

Dadeos-Menlo: No worries. I greatly appreciate your help and insight, not to mention the general implementation of "cloud-provider-kind".

There's no rush on my account - I have a working system now - and it will take me a while to sort the CLA.

If you wish to abandon this pull-request and re-implement it yourself, then that would be find with me - you did most of the work, I just did some typing.

BenTheElder: > @BenTheElder you have any idea on how to identify if the containers are running inside a VM as in Mac or Windows?

Doing this reliably from within the container is problematic, there are things we can sniff but nothing guaranteed.
Instead probably detecting the container itself from the docker / nerdctl API could work, but I don't think we can support this for podman.

> I'm not sure that that is a valid distinction. The system either runs within a container or it doesn't; whether the container runtime is hosted directly on the host or within a VM running on the host is not really its concern?

The networking and filesystem implications are substantially different, for starters. When using something like docker desktop there is an expectation to bridge the boundary between the host and the containers in the VM, we're a lot more limited in our options if we're running entirely within the VM.

It's true that we use docker portmap currently though.

> Indeed; but - unless people go to the trouble of defining a systemd unit, or other automated launch mechanism, or manually launch the "cloud-provider-kind" executable every time they want to use it - why wouldn't everyone want to ensure that the system was running whenever their Kind cluster[s] are running (i.e. whenever the Docker daemon is running, which may be trivially achieved by running it within a container using --restart=always)?

This isn't necessarily possible in podman and nerdctl, which we also support[^1].

We could provide a reference systemd unit, however currently the expectation is that you'll run this in the background while you develop.

[^1]: sort of, it's not that simple, but in broad strokes this is a rather docker specific thing last I checked, podman and nerdctl do not have the same restart functionality.

pkg/controller/controller.go
```diff
@@ -111,14 +111,13 @@ func (c *Controller) Run(ctx context.Context) {
 // getKubeClient returns a kubeclient depending if the ccm runs inside a container
 // inside the same docker network that the kind cluster or run externally in the host
 // It tries first to connect to the external endpoint
-func (c *Controller) getKubeClient(ctx context.Context, cluster string) (kubernetes.Interface, error) {
+func (c *Controller) getKubeClient(ctx context.Context, cluster string) (kubernetes.Interface, bool, error) {
```
BenTheElder, line 114: this should maybe be an "enum", there's host, in-network IP, in-network container name/DNS, etc ...

it also feels like a bit of a conflation of concerns to return a kubernetes client AND the network connectivity mode here

aojea:
> For example; I've defined a Cilium [Gateway](https://docs.cilium.io/en/stable/network/servicemesh/gateway-api/gateway-api/#gateway-api-support) bound to port `14000`, which I am able to access from the host (i.e. my Mac) via the port mapped by Docker (i.e. `0.0.0.0:32774->14000/tcp`):

@Dadeos-Menlo one question, so when you create the Service , as per example

```
kubectl get service/lb-service-local
NAME               TYPE           CLUSTER-IP      EXTERNAL-IP   PORT(S)        AGE
lb-service-local   LoadBalancer   10.96.207.137   192.168.8.7   80:31215/TCP   57s
```

the tunnelManager what it does is to expose the Service Port, in this case 80 in the host.

When running the kindccm inside a container in mac or windows, the tunnel manager can not work, because it runs in the VM, and does not have access to the host.

IIUIC what you are using is only the portmap capability of the tunnel manager, right?

Dadeos-Menlo:
Running the [examples/loadbalancer_etp_local.yaml](https://github.com/kubernetes-sigs/cloud-provider-kind/blob/f26b241ce573c53e81dfca8070167b0288626c03/examples/loadbalancer_etp_local.yaml#L1) with the changes, as currently [proposed](https://github.com/kubernetes-sigs/cloud-provider-kind/pull/125/commits/5e03531c9674276a444e7270a1452bfc3c03a048), yields:
```
kubectl get service/lb-service-local                          
NAME               TYPE           CLUSTER-IP     EXTERNAL-IP   PORT(S)        AGE
lb-service-local   LoadBalancer   10.96.205.97   <pending>     80:30236/TCP   11s
```
I was mistaken as to the requirement/significance of the `tunnelManager` implementation due to the fact that the required ports are _only_ published by the proxy container when the `tunnelManager` exists: https://github.com/kubernetes-sigs/cloud-provider-kind/blob/f26b241ce573c53e81dfca8070167b0288626c03/pkg/loadbalancer/server.go#L231-L239

the `tunnelManager` itself isn't working in my scenario, which I imagine is to be expected, hence the "EXTERNAL-IP" being shown as "\<pending\>".

I _suspect_ that a more appropriate fix for the lack of publishing of ports is to publish them `if s.directConnectivity || s.tunnelManager != nil`, rather than attempting to instantiate the `tunnelManager` in the `directConnectivity` case. But I don't understand why one wouldn't simply always publish the ports from the proxy container; like the `envoyAdminPort` is.

aojea:
> or I can access it from the host via the ports published via Docker.

The goal of the project is to provide the same user experience than a cloud provider and its loadbalancers, however, it also covers this gap of dynamic access to Services, and this scenario of the ccm running in a container in a Mac environment is something that only happens on this setup.

What you want to have is dynamic portmapping for the Service. Pormapping is an internal detail to provide the tunneling on Mac and Windows, always enabling portmaps waste resources.

What we have is these 4 scenarios:

1. Containers Reachable from host and KCCM runs as a Process: Direct Connectivity (Linux)
   - Same experience as a real loadbalancer
2. Containers Reachable from host and KCCM runs as a container: Direct Connectivity (Linux)
   - Same experience as a real loadbalancer
3. Containers Not Reachable  from host and KCCM runs as a Process: Tunnel Connectivity (Mac, Windows, WSL2) 
   - Same experience as a real loadbalancer
4. Containers Not Reachable  from host and KCCM runs as a container: Portmap (Mac, Windows, WSL2)
   - This is your scenario, this is just a workaround to provide connectivity

The heuristic here is not "direct connectivity" or we'll be enabling pormapping on the Linux environments that is totally undesirable.

The heuristic is "Host is Mac/Windows and KCCM runs in a container, where runs in a container can be replaced by has direct connectivity with the containers if we know the host is mac or windows

Dadeos-Menlo:
> The goal of the project is to provide the same user experience than a cloud provider and its loadbalancers, however, it also covers this gap of dynamic access to Services, and this scenario of the ccm running in a container in a Mac environment is something that only happens on this setup.

I must confess that I am not entirely familiar with every aspect of the load-balancer functionality offered by every cloud provider. I imagine that an "EXTERNAL-IP" is associated with the provisioned load-balancer and that this IP address is discoverable via interrogation of the Kubernetes API. I presume that, generally, such IP addresses are accessible via the Internet, but I can also imagine scenarios where a load-balancer only accessible from a private network might also be provisioned (i.e. "external" in this instance is likely in contrast to the _internal_ cluster network).

> What do you want is to have dynamic portmapping for the Service. Portmapping is an internal detail to provide the tunneling on Mac and Windows, always enabling portmaps waste resources, however, it seems you find it useful and probable other users.

What I'm really after is a configuration that _works_ in my development environment (i.e. on a Mac). I suspect that this will involve usage of ports "published" by the proxy container via Docker; but that's predominantly because that's my understanding of how Docker works. Take, by way of an example, accessing the API server of a Kind cluster (i.e. port 6443), there are two addresses that might be used (i.e. whether `internal` is "true" or "false") one is accessing port 6443 on the IP address of the "…-control-plane" container (i.e. on the "kind" network) and the other is accessing whatever port port 6443 has been mapped to on the host system via Docker's port publishing mechanism.

_Normally_, when accessing a Kubernetes API server, one would simply access port 6443. However, when dealing with the special case of running Kubernetes on Docker (i.e. Kind) it seems reasonable to assume a degree of appreciation by the user of some of the additional considerations introduced by the use of Docker (i.e. the "kind" virtual network and the potential use of published ports).

> What we have is these 4 scenarios:
> 
> 1. Containers Reachable from host and KCCM runs as a Process: Direct Connectivity (Linux)
>    
>    * Same experience as a real loadbalancer
> 2. Containers Reachable from host and KCCM runs as a container: Direct Connectivity (Linux)
>    
>    * Same experience as a real loadbalancer
> 3. Containers Not Reachable  from host and KCCM runs as a Process: Tunnel Connectivity (Mac, Windows, WSL2)
>    
>    * Same experience as a real loadbalancer
> 4. Containers Not Reachable  from host and KCCM runs as a container: Portmap (Mac, Windows, WSL2)
>    
>    * This is your scenario, this is just a workaround to provide connectivity

I don't entirely agree with the scenario classifications as identified - although entirely accept that I am no expert in this field. I suggest that there are 2 variants of scenario 2, running "cloud-provider-kind" within a container on Linux, depend whether `--network host` or `--network kind` is used. None of the supposed tunnel connectivity purported to be associated with scenarios 3 and 4 work for me. I doubt whether it's possible to achieve the intended behaviour on non-Linux deployments - where I presume the intended behaviour is to have some magic IP address that is directly accessible from the host (e.g. a Mac).

I am not wedded to the idea of running "cloud-provider-kind" within a container; although I don't understand why anyone wouldn't, because running it directly on my Mac seems to offer the same results as running it in a container.

> 
> The heuristic here is not "direct connectivity" or we'll be enabling pormapping on the Linux environments that is totally undesirable.
> 
> The heuristic is "Host is Mac/Windows and KCCM runs in a container, where runs in a container can be replaced by has direct connectivity with the containers if we know the host is mac or windows ...
> 
> Being able to know if is mac or windows from inside the container will be ideal, otherwise we can set a magic env variable or a flag

This doesn't seem correct to me. I can imagine that there might be a special case of running on Linux using "host" networking (i.e. either by running "cloud-provider-kind" natively on the Linux box or within a container with `--network host`) whereby it's possible to arrange for the load balancer to have an IP address that can simply be accessed from the host network. But normal usage would seem to be to consider the load balancer to be part of the "kind" network, and I don't believe that whether the system is running on Linux, Mac, or Windows, should affect the behaviour and interaction in such a deployment scenario.

aojea:
> why anyone wouldn't, because running it directly on my Mac seems to offer the same results as running it in a container.

```
kubectl get service/lb-service-local 
NAME               TYPE           CLUSTER-IP     EXTERNAL-IP   PORT(S)        AGE
lb-service-local   LoadBalancer   10.96.205.97   172.18.0.4    80:30236/TCP   37m
```

when running in Mac in the host with this example you do `curl http://172.18.0.4`  in Mac and you reach the exposed service.

If you run `cloud-provider-kind` in Mac in a container, you need to figure out the portmapping , don't you?

Kubernetes loadbalancers work by connecting the ExternalIP on the same port the service is exposed.

aojea:
@Dadeos-Menlo I merged the other PR and added you as coauthor of the commit, so I'm going to close this one

Thanks for your feedback and help

/close

PR #129, opened 2024-09-12
Update go.mod

cloudikeme:
Bumped up go version from 1.22 to 1.23


aojea:
Thanks, but it is better we keep compatibility with 1.22, go.mod 

This statement is usually confused https://go.dev/doc/modules/gomod-ref#go

Issue #144, opened 2024-10-22
command(s?) in documentation don't work. 


mikedlr:
In the main documentation there is a command

    $ kubectl label node kind-control-plane node.kubernetes.io/exclude-from-external-load-balancers-
    node/kind-control-plane unlabeled

This command does not work. I believe that it should be something like

    kubectl label node --overwrite kind-2-control-plane node.kubernetes.io/exclude-from-external-load-balancers=true

instead - probably better if the prompt isn't included in the copied text.  It might be good if the documents were automatically tested e.g. with https://github.com/JNRowe/shell-doctest

BenTheElder:
> It might be good if the documents were automatically tested e.g. with https://github.com/JNRowe/shell-doctest

that might get a bit expensive since you'd need the requisites setup.

they're also probably not changing often.

I agree that we should have copyable commands.

zrisher:
I also personally felt the documentation around this command was unclear:

> Control-plane nodes need to remove the special label node.kubernetes.io/exclude-from-external-load-balancers to be able to access the workloads running on those nodes using a LoadBalancer Service.

Does "these nodes" mean just the control-plane node? Or the nodes it controls? I.e. do I need to run this command if I don't plan on putting a load balancer in front of my control plane?

aojea:
@zrisher how familiar are you with kubernetes? just trying to understand if you are looking at this from a totally new person or with some background

zrisher:
First K8S project, a lot of hours in but also many new technologies to research besides just K8S. Definitely the first time I've tried to get a load balancer running on a local cluster.

aojea:
yeah, I see, this project README assumes some previous knowledge, now I understand why you feel this friction

zrisher:
If this configuration step is needed in all cases, I think

> WE also need to remove the label `node.kubernetes.io/exclude-from-external-load-balancers` from the control plane node:

is more clear for both the experienced and the iniates.

BenTheElder:
FWIW I personally think it's generally best practice that _examples_ have `$` and output and "you should run this" only have exactly the command to copy/paste and I would support a PR to correct that.

---

> Does "these nodes" mean just the control-plane node? Or the nodes it controls? I.e. do I need to run this command if I don't plan on putting a load balancer in front of my control plane? Or is this something I need to do to use this tool at all?

There can be more than one control-plane node, which is why it says "these nodes". The nodes with this label.
That label controls excluding nodes from load balancing.

[...]

PR #145, opened 2024-10-22
Make labelling command that doesn't fail in README.md

mikedlr:
Designed as a fix for https://github.com/kubernetes-sigs/cloud-provider-kind/issues/144 this gives a labeling command that completes and *might* do the expected job. "It works for me"(TM)

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,8 +90,7 @@ Have a question, bug, or feature request? Let us know! https://kind.sigs.k8s.io/
 Control-plane nodes need to remove the special label `node.kubernetes.io/exclude-from-external-load-balancers` to be able to access the workloads running on those nodes using a LoadBalancer Service.
 
 ```sh
-$ kubectl label node kind-control-plane node.kubernetes.io/exclude-from-external-load-balancers-
-node/kind-control-plane unlabeled
+kubectl label node kind-control-plane node.kubernetes.io/exclude-from-external-load-balancers=true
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -90,8 +90,7 @@ Have a question, bug, or feature request? Let us know! https://kind.sigs.k8s.io/
 Control-plane nodes need to remove the special label `node.kubernetes.io/exclude-from-external-load-balancers` to be able to access the workloads running on those nodes using a LoadBalancer Service.
 
 ```sh
-$ kubectl label node kind-control-plane node.kubernetes.io/exclude-from-external-load-balancers-
-node/kind-control-plane unlabeled
+kubectl label node kind-control-plane node.kubernetes.io/exclude-from-external-load-balancers=true
```
aojea, line 93:
```suggestion
kubectl label node kind-control-plane node.kubernetes.io/exclude-from-external-load-balancers-
```

the goal is to remove the label not to set it to true, no?

mikedlr: yes, it seems you are right that was the opposite of what I understood because it was unlabelled when I got a failed command but I think you are right. It either was never labelled or (more likely) I managed to unlabel it without realizing.

mikedlr:
Think this was wrong and the change to make the commands work on cut and paste is bigger so closing this and I'll think about if I understand enough to do the other.

PR #156, opened 2024-11-12
fix makefile to override image name

upodroid:
#155 didn't work as the makefile is ignoring the env override.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,9 @@ can make it available elsewhere if appropriate:
 sudo install ~/go/bin/cloud-provider-kind /usr/local/bin
 ```
 
+Starting with v0.4.0, the docker image for cloud-provider-kind is available
+at `registry.k8s.io/cloud-provider-kind/cloud-controller-manager`
+
 You can also build it locally:
 
 ```sh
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -29,6 +29,9 @@ can make it available elsewhere if appropriate:
 sudo install ~/go/bin/cloud-provider-kind /usr/local/bin
 ```
 
+Starting with v0.4.0, the docker image for cloud-provider-kind is available
+at `registry.k8s.io/cloud-provider-kind/cloud-controller-manager:v0.4.0`
```
stmcginnis, line 33: I wonder if we should leave off the tag, or default it to `latest`. We'll either have to remember to update the README for every release, or we'll have a risk of people copy/pasting this without thinking and being stuck using what will eventually be an old release.

stmcginnis (commented):
Good catch!

/lgtm

README.md
upodroid: Its fixed, can you also edit this release as well? https://github.com/kubernetes-sigs/cloud-provider-kind/releases/tag/v0.4.0 the last line in the release note

aojea:
/approve

Thanks

README.md
aojea: done

stmcginnis:
/lgtm

Thanks!

Issue #171, opened 2024-12-11
flaky image builds


BenTheElder:
https://testgrid.k8s.io/sig-testing-kind#post-cloud-provider-kind-image

aojea:
> Your build timed out. Use the [--timeout=DURATION] flag to change the timeout threshold.

how can it take more than one hour https://prow.k8s.io/job-history/gs/kubernetes-ci-logs/logs/post-cloud-provider-kind-image

the good one takes 5 mins

stmcginnis:
Unfortunately it doesn't have timestamps for most of it, but some areas that could cause delays that come to mind:

- Throttling during pulling docker image layers
- Resource contention on the build host slowing down the process
- Throttling during pulling go packages for the build

We could probably add some explicit timestamp logging to help narrow down what areas are taking so long. Just trying to brainstorm ideas.

BenTheElder:
or qemu flakes? we are compiling under emulation instead of cross compiling, we should fix that. it will be faster and more reliable.

BenTheElder:
I'll send a cross-compile PR for the docker image ...

BenTheElder:
the push job is substantially faster now after https://github.com/kubernetes-sigs/cloud-provider-kind/pull/173 as expected, reliably under 3m.

BenTheElder:
previously ~5m-9m with the occasionally 1h+ hang:
https://prow.k8s.io/job-history/gs/kubernetes-ci-logs/logs/post-cloud-provider-kind-image 

I'm going to presume this is closed for now and reopen it later if we get another alert for flakes

PR #173, opened 2024-12-12
docker: cross compile binary

BenTheElder:
see discussion in https://github.com/kubernetes-sigs/cloud-provider-kind/issues/171, not sure if this is a complete fix but it should help, we avoid qemu entirely

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,12 @@
-FROM golang:1.22
+FROM --platform=$BUILDPLATFORM golang:1.22
 WORKDIR /go/src
 # make deps fetching cacheable
 COPY go.mod go.sum ./
 RUN go mod download
 # build
 COPY . .
-RUN make build
+ARG TARGETARCH
+RUN GOARCH=$TARGETARCH make build
 
 # build real cloud-provider-kind image
 FROM docker:25.0-dind


Dockerfile
```diff
@@ -1,11 +1,12 @@
...
 COPY go.mod go.sum ./
 RUN go mod download
 # build
 COPY . .
-RUN make build
+ARG TARGETARCH
+RUN GOARCH=$TARGETARCH make build
 
 # build real cloud-provider-kind image
 FROM docker:25.0-dind
```
BenTheElder, line 12: will make a seperate PR, but we have no reason to actually ship the daemon right? this image is bigger than we need, we just need the `docker` CLI, right? or am I missing something.

aojea: we do, the cloud provider kind needs docker when running inside docker

BenTheElder: confirmed we only need the CLI with @aojea

BenTheElder: filed as another PR so we can decide separately https://github.com/kubernetes-sigs/cloud-provider-kind/pull/174, but included this existing commit to avoid diff since it's about to merge hopefully

BenTheElder:
> pull-cloud-provider-kind-conformance-parallel-ga-only

these jobs are not blocking merge

probably we should run just a subset of loadbalancer related tests and require it for merge?

Issue #185, opened 2025-01-13
Improve startup latency


aojea:
I think we can improve that, right now is a very basic serial check

https://github.com/kubernetes-sigs/cloud-provider-kind/blob/0ca70805fdb846aacae28c66d75b4dd0dd287d0b/pkg/controller/controller.go#L121-L171

we can run both checks in parallel and choose the one that wins the race, I will open a feature request to fix this , otherwise is hard to follow this issue if we have multiple topics

_Originally posted by @aojea in https://github.com/kubernetes-sigs/cloud-provider-kind/issues/184#issuecomment-2586867946_

PR #187, opened 2025-01-15
Improve startup latency

aojea:
Improve the cloud provider kind startup latency by polling the endpoints of the apiserver in parallel instead of serial

Fixes: https://github.com/kubernetes-sigs/cloud-provider-kind/issues/185

FYI @slartibart70


stmcginnis:
Nice!

/lgtm

Issue #190, opened 2025-01-17
Use a shorter hash for the container name


aojea:
> I0117 14:46:26.851473   32108 proxy.go:236] address type Hostname, only InternalIP supported
> 
> it seems is not able to obtain the ip of the container, try
> 
> `docker inspect -f {{range .NetworkSettings.Networks}}{{.IPAddress}},{{.GlobalIPv6Address}}{{end}} kindccm-RZTQCN5REF5VESVEBSRFF5NKA2S54JVJFWGT3EOM` , with kindccm-... the current container (we should replace such long string for something shorter) 

 _Originally posted by @aojea in [#189](https://github.com/kubernetes-sigs/cloud-provider-kind/issues/189#issuecomment-2597941065)_

PR #191, opened 2025-01-19
Use a shorter hash for the container name

priyanshikhetwani:
`docker inspect -f {{range .NetworkSettings.Networks}}{{.IPAddress}},{{.GlobalIPv6Address}}{{end}} kindccm-RZTQCN5REF5VESVEBSRFF5NKA2S54JVJFWGT3EOM , with kindccm-... the current container (we should replace such long string for something shorter)`

Fixes: #190

### DIFF
--- a/pkg/loadbalancer/server.go
+++ b/pkg/loadbalancer/server.go
@@ -163,7 +163,7 @@ func (s *Server) EnsureLoadBalancerDeleted(ctx context.Context, clusterName stri
 func loadBalancerName(clusterName string, service *v1.Service) string {
 	hash := sha256.Sum256([]byte(loadBalancerSimpleName(clusterName, service)))
 	encoded := base32.StdEncoding.EncodeToString(hash[:])
-	name := constants.ContainerPrefix + "-" + encoded[:40]
+	name := constants.ContainerPrefix + "-" + encoded[:16]
 
 	return name
 }
--- a/pkg/loadbalancer/server.go
+++ b/pkg/loadbalancer/server.go
@@ -162,7 +162,10 @@ func (s *Server) EnsureLoadBalancerDeleted(ctx context.Context, clusterName stri
 // loadbalancer name is a unique name for the loadbalancer container
 func loadBalancerName(clusterName string, service *v1.Service) string {
 	h := sha256.New()
-	io.WriteString(h, loadBalancerSimpleName(clusterName, service))
+	_, err := io.WriteString(h, loadBalancerSimpleName(clusterName, service))
+	if err != nil {
+		panic(err)
+	}
 	hash := h.Sum(nil)
 	return fmt.Sprintf("%s-%x", constants.ContainerPrefix, hash[:6])
 }
--- a/pkg/loadbalancer/server.go
+++ b/pkg/loadbalancer/server.go
@@ -3,9 +3,9 @@ package loadbalancer
 import (
 	"context"
 	"crypto/sha256"
-	"encoding/base32"
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"path"
 	"strings"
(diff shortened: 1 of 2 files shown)


pkg/loadbalancer/server.go
```diff
@@ -163,7 +163,7 @@ func (s *Server) EnsureLoadBalancerDeleted(ctx context.Context, clusterName stri
 func loadBalancerName(clusterName string, service *v1.Service) string {
 	hash := sha256.Sum256([]byte(loadBalancerSimpleName(clusterName, service)))
 	encoded := base32.StdEncoding.EncodeToString(hash[:])
-	name := constants.ContainerPrefix + "-" + encoded[:40]
```
aojea, line 166: let's print the first 6 hash bytes that is more than enough to avoid collisions on names 
https://en.wikipedia.org/wiki/Birthday_problem

See https://en.wikipedia.org/wiki/Birthday_problem

```go
        h := sha256.New()
	io.WriteString(h, loadBalancerSimpleName(clusterName, service))
	hash := h.Sum(nil)
	return fmt.Sprintf("%s-%x", constants.ContainerPrefix, hash[:6])
```

aojea: you also need to update the test in with the new name

https://github.com/kubernetes-sigs/cloud-provider-kind/blob/bdc116dd8bf7cb3c9dbc69fd8fb83974c6c2d321/pkg/loadbalancer/server_test.go#L11-L38

pkg/loadbalancer/server.go
```diff
@@ -161,11 +161,10 @@ func (s *Server) EnsureLoadBalancerDeleted(ctx context.Context, clusterName stri
 
 // loadbalancer name is a unique name for the loadbalancer container
 func loadBalancerName(clusterName string, service *v1.Service) string {
-	hash := sha256.Sum256([]byte(loadBalancerSimpleName(clusterName, service)))
-	encoded := base32.StdEncoding.EncodeToString(hash[:])
-	name := constants.ContainerPrefix + "-" + encoded[:40]
-
-	return name
+	h := sha256.New()
+	io.WriteString(h, loadBalancerSimpleName(clusterName, service))
```
aojea, line 165:
```suggestion
	_ = io.WriteString(h, loadBalancerSimpleName(clusterName, service))
```
to avoid the linter error

pkg/loadbalancer/server.go
```diff
@@ -161,11 +161,13 @@ func (s *Server) EnsureLoadBalancerDeleted(ctx context.Context, clusterName stri
...
 func loadBalancerName(clusterName string, service *v1.Service) string {
-	hash := sha256.Sum256([]byte(loadBalancerSimpleName(clusterName, service)))
-	encoded := base32.StdEncoding.EncodeToString(hash[:])
-	name := constants.ContainerPrefix + "-" + encoded[:40]
-
-	return name
+	h := sha256.New()
+	_, err := io.WriteString(h, loadBalancerSimpleName(clusterName, service))
+	if err != nil {
+		panic(err)
```
aojea, line 167: it seems this never return an error https://cs.opensource.google/go/go/+/refs/tags/go1.23.5:src/crypto/sha256/sha256.go;l=174-195

pkg/loadbalancer/server_test.go
```diff
@@ -20,8 +20,8 @@ func TestLoadBalancerName(t *testing.T) {
 			name:        "simple",
 			cluster:     "test-cluster",
 			service:     &v1.Service{ObjectMeta: metav1.ObjectMeta{Namespace: "test-namespace", Name: "test-service"}},
-			expected:    constants.ContainerPrefix + "-CGVXJAVBASN2Z3RXOABMYVHNP7WNHR3ATSDVOTEN",
```
aojea, line 23: a cloud-provider-kind delete the old loadbalancers on exit and create new ones, so this change should not break existing users

aojea:
/lgtm
/approve

Thanks

Issue #192, opened 2025-01-20
Validate privileges in Windows environments


aojea:
> Thank you so much! I forgot to enable privileged mode in Windows and WSL2. Now it works properly.
> 
> I am using Docker Desktop shared between WSL2 and Windows, and I’ve tried both NAT and mirrored WSL2 network configurations.
> 
> Notably:
> 
> - When using NAT mode, cloud-provider-kind needs to be started in WSL2 with privileged mode, and the service can only be accessed via curl from WSL2, not from Windows.
> 
> - When using mirrored mode, cloud-provider-kind needs to be started in Windows with privileged mode, and the service can only be accessed via curl from Windows, not from WSL2.
> 
> Both configurations can assign an external IP to the LoadBalancer service. However, is there a way to make the LoadBalancer service accessible from both WSL2 and Windows? 

 _Originally posted by @d2461795341 in [#189](https://github.com/kubernetes-sigs/cloud-provider-kind/issues/189#issuecomment-2598364846)_

It will be good to add a check for windows environments to validate there is enough privileges or fail , same as we do for Mac

https://github.com/kubernetes-sigs/cloud-provider-kind/blob/bdc116dd8bf7cb3c9dbc69fd8fb83974c6c2d321/cmd/app.go#L74-L77

dprotaso:
I just hit this :)

dprotaso:
Am I supposed to run cloud-provider-kind with `sudo` ? and it should work

dprotaso:
Reading the subsequent comment in the linked thread - it requires `sudo` and `--enable-lb-port-mapping=true`

dprotaso:
Though I'm unable to curl the Service LB IP - it doesn't seem to work hmm.

aojea:
> Though I'm unable to curl the Service LB IP - it doesn't seem to work hmm.

you should not need to enable-lb-portmapping, that is a flag for environments that does not allow to create direct routing to the containers so people can still find the portmaps in the host and use them, but was added just as a special option for these edge cases  https://github.com/kubernetes-sigs/cloud-provider-kind/pull/126

If you want to use the Service LB IP you can not set that flag

https://github.com/kubernetes-sigs/cloud-provider-kind/blob/134f5bed73ee4889237c169e7cd6048717859ce2/cmd/app.go#L136-L144

dprotaso:
> If you want to use the Service LB IP you can not set that flag

[...]

PR #205, opened 2025-01-30
If the IP is on the lo device already don't fail

dprotaso:
This might be the fix for https://github.com/kubernetes-sigs/cloud-provider-kind/issues/189 or https://github.com/kubernetes-sigs/cloud-provider-kind/issues/192

### DIFF
--- a/pkg/loadbalancer/server.go
+++ b/pkg/loadbalancer/server.go
@@ -121,7 +121,7 @@ func (s *Server) EnsureLoadBalancer(ctx context.Context, clusterName string, ser
 		klog.V(2).Infof("updating loadbalancer tunnels on userspace")
 		err = s.tunnelManager.setupTunnels(loadBalancerName(clusterName, service))
 		if err != nil {
-			klog.V(2).ErrorS(err, "error setting up tunnels")
+			klog.ErrorS(err, "error setting up tunnels")
 		}
 	}
 
--- a/pkg/loadbalancer/server.go
+++ b/pkg/loadbalancer/server.go
@@ -121,7 +121,7 @@ func (s *Server) EnsureLoadBalancer(ctx context.Context, clusterName string, ser
 		klog.V(2).Infof("updating loadbalancer tunnels on userspace")
 		err = s.tunnelManager.setupTunnels(loadBalancerName(clusterName, service))
 		if err != nil {
-			return nil, err
+			klog.V(2).ErrorS(err, "error setting up tunnels")
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


aojea:
great catch, I think we may keep that the same and instead, on the caller, just ignore the error and log, but not stop the loadbalancer creation, since the tunnels may have different errors on different OS I think this approach is more resilient

https://github.com/kubernetes-sigs/cloud-provider-kind/blob/defae812a313ae5551727869139734903852da36/pkg/loadbalancer/server.go#L120-L126

replacing `return nil, err `by a log `klog.Infof("Failed to create tunn: %v", err)`

pkg/loadbalancer/tunnel.go
```diff
@@ -39,9 +39,9 @@ func (t *tunnelManager) setupTunnels(containerName string) error {
 	}
 
 	klog.V(0).Infof("setting IPv4 address %s associated to container %s", ipv4, containerName)
-	err = AddIPToLocalInterface(ipv4)
+	output, err := AddIPToLocalInterface(ipv4)
 	if err != nil {
-		return err
+		klog.V(4).Infof("error adding IP to local interface: %v - %s", err, output)
```
aojea, line 44: wrap the error here, I think this `tunnel` is ok to return the error ... I find better if we swallow it here by logging it at normal level, as the tunnel is an addon for certain environments https://github.com/kubernetes-sigs/cloud-provider-kind/blob/defae812a313ae5551727869139734903852da36/pkg/loadbalancer/server.go#L120-L126

dprotaso: pushed - take a look at the edits

pkg/loadbalancer/server.go
```diff
@@ -121,7 +121,7 @@ func (s *Server) EnsureLoadBalancer(ctx context.Context, clusterName string, ser
 		klog.V(2).Infof("updating loadbalancer tunnels on userspace")
 		err = s.tunnelManager.setupTunnels(loadBalancerName(clusterName, service))
 		if err != nil {
-			return nil, err
+			klog.V(2).ErrorS(err, "error setting up tunnels")
```
aojea, line 124:
```suggestion
			klog.ErrorS(err, "error setting up tunnels")
```

last thing and we are good, I prefer to log this always for less advanced users

aojea:
/approve

one change only for final lgtm  https://github.com/kubernetes-sigs/cloud-provider-kind/pull/205#discussion_r1936194329

Thanks

aojea:
/lgtm
/approve

Thanks

PR #207, opened 2025-01-31
Fix a command that removes a node in README.md

makotonakai:
**About**
This PR fixes a command that excludes a particular worker node from the list of backend servers.

**Reference:**
https://kubernetes.io/docs/reference/labels-annotations-taints/#node-kubernetes-io-exclude-from-external-load-balancers

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ If you are running workloads on control plane nodes, as is the [default kind con
 you will need to remove this label to access them using a LoadBalancer:
 
 ```sh
-$ kubectl label node kind-control-plane node.kubernetes.io/exclude-from-external-load-balancers-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -97,7 +97,7 @@ If you are running workloads on control plane nodes, as is the [default kind con
 you will need to remove this label to access them using a LoadBalancer:
 
 ```sh
-$ kubectl label node kind-control-plane node.kubernetes.io/exclude-from-external-load-balancers-
```
aojea, line 100: this is on purpose, you need to remove this label

makotonakai: Sorry, I just realized that I totally missed the point.
Let me close this PR.

PR #227, opened 2025-04-14
support Gateway API (alpha) Gateway and HTTPRoute only

aojea:
This is a basic implementation of Gateway API done with the goal of having something that passes conformance and we can iterate. Code is not great, I just focused on get it working and Gateway API is complex, really complex, and envoy is too and my knowledge is very limited ... the bidirectional relations between Listeners and Routes have a lot edge cases that conformance test happily cover (at least it seems to cover a considerable number , Kudos to the community, great job)

It uses envoy because is where we settled with the implementation of the loadbalancers, no other reasons than it has a more flexible API than what we had with nginx or haproxy where we relied on templating ... I do nto feel templating is the right way to program dataplanes, see I also moved to pure netlink communication for all nftables stuff in the other projects.

The goals are :
- let people play with gateway , cloud-provider-kind is already widely for training, CI, development, testing .... there is a big barrier and steep curve , hope this lower it
- keep iterating and harden the implementation, also server as developer tool for the community, so we can also test upgrades and migrations or switches between experimental/standard ...

Happy hacking and please provide feedback ... and if you have time try to get more coverage in the rest of gateway api features.


pkg/gateway/controller.go
```diff
@@ -0,0 +1,203 @@
...
+			}
+		},
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	_, err = httprouteInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
```
toVersus, line 111: Just like how GatewayClass is used to filter in the Gateway's event handler, HTTPRoutes and GRPCRoutes should also be reconciled only for resources associated with their Gateway (GatewayClass). Otherwise, multiple Gateway API controllers could end up operating on the same resource.

Is it common to look at the [ParentRef](https://gateway-api.sigs.k8s.io/reference/spec/#parentreference) in HTTPRoute or GRPCRoute and then check the GatewayClass of the referenced Gateway?

aojea: added it now, PTAL

go.mod
```diff
@@ -63,7 +73,8 @@ require (
 	golang.org/x/term v0.32.0 // indirect
 	golang.org/x/text v0.25.0 // indirect
 	golang.org/x/time v0.11.0 // indirect
-	google.golang.org/protobuf v1.36.6 // indirect
+	google.golang.org/genproto/googleapis/api v0.0.0-20250106144421-5f5ef82da422 // indirect
```
BenTheElder, line 76: why is this being pulled in?

go.mod
```diff
@@ -17,17 +21,22 @@ require (
 	k8s.io/controller-manager v0.33.0
 	k8s.io/klog/v2 v2.130.1
 	k8s.io/utils v0.0.0-20250502105355-0f33e8f1c979
+	sigs.k8s.io/gateway-api v1.3.0
```
BenTheElder, line 24: it would be really nice if we could import just the API without bringing dependencies like DNS, controller-runtime, etc into our dependency graph

pkg/gateway/routes.go
```diff
@@ -0,0 +1,231 @@
...
+func getFirstKey[K comparable, V any](m map[K]V) K {
+	for k := range m {
+		return k
+	}
+	var zero K
+	return zero
+}
+
+// toInterfaceSlice converts a slice of a specific type to a slice of interface{}.
+func toInterfaceSlice[T any](s []*T) []interface{} {
```
BenTheElder, line 225: seems inconsistent to take `any` and then return `[]interface{}`, `any` is an alias for`interface{}`

.github/workflows/gateway.yml
```diff
@@ -0,0 +1,137 @@
...
+            --debug=true \
+            --kubeconfig=_artifacts/kubeconfig.conf \
+            --organization=sigs.k8s.io \
+            --project=cloud-provider-kind \
+            --url=https://github.com/kubernetes-sigs/cloud-provider-kind \
+            --version=test \
+            --contact=antonio.ojea.garcia@gmail.com \
+            --gateway-class=cloud-provider-kind \
+            --supported-features=Gateway,HTTPRoute \
+            --report-output=./_artifacts/conformance.yaml
```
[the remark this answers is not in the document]
LiorLieberman: Run with 
1. remove `--supported-features=Gateway,HTTPRoute `
2. add  `--conformance-profiles=GATEWAY-HTTP --all-features`
3. add -v to get output

PR #229, opened 2025-04-17
make loadbalancer tunnel protocol aware

aojea:
We were only tunneling TCP but seems UDP is also a requested feature

### DIFF
--- a/pkg/container/container.go
+++ b/pkg/container/container.go
@@ -220,7 +220,7 @@ func PortMaps(name string) (map[string]string, error) {
 		// TODO we just can get the first entry or look for ip families
 		for _, pm := range v {
 			if pm.HostPort != "" {
-				result[parts[0]] = pm.HostPort
+				result[parts[0]+"/"+protocol] = pm.HostPort
 				break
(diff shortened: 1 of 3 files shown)


aojea:
@stmcginnis lol, I didn't tested, I do not know if it works :)

stmcginnis:
Doh, I guess I should have asked first!

Should we revert? Or do you want to test and follow up with any corrections?

aojea:
> Should we revert? Or do you want to test and follow up with any corrections?

do you have. mac to test this?

I'm trying to see if I can add automated testing in https://github.com/kubernetes-sigs/cloud-provider-kind/pull/230

stmcginnis:
Unfortunately no, only access to Linux environments at the moment.

BenTheElder:
I forget: can we tunnel all? What about our friend SCTP?

aojea:
> I forget: can we tunnel all? What about our friend SCTP?

That's not used out of very niche cases, I will be surprised someone uses it in Mac or windows, or if those platforms even support it

PR #230, opened 2025-04-17
add bats tests and fix udp on mac and windows

aojea:
Allow to forward UDP on Mac and Windows

Co-author-by: @imchristianlowe

### DIFF
--- a/.github/workflows/bats.yml
+++ b/.github/workflows/bats.yml
@@ -1,0 +1,77 @@
+name: bats
+
+on:
+  push:
+    branches:
+      - 'main'
+    tags:
+      - 'v*'
+  pull_request:
+    branches: [ main ]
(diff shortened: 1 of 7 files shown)


BenTheElder (commented):
wait, are we using external bash libraries? wheee

.github/workflows/bats.yml
```diff
@@ -0,0 +1,74 @@
...
+  K8S_VERSION: "v1.32.0"
+  KIND_VERSION: "v0.27.0"
+
+jobs:
+  bats_tests:
+    runs-on: ubuntu-22.04
+    name: Bats e2e tests
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
```
BenTheElder, line 24: pin these (commit hash, see other flows)

PR #241, opened 2025-05-14
remove vendor folder and update deps

aojea:
Change-Id: I35e3a8081da2a486c2ae6be2c9a64fc00bfd7186


BenTheElder:
I've actually been debating if we need to add it to other repos, it's hard to track dependency changes :/
And the module proxy doesn't guarantee availability of content

aojea:
> I've actually been debating if we need to add it to other repos, it's hard to track dependency changes :/ And the module proxy doesn't guarantee availability of content

I do not want to carry +4.5k files in the repo for a project like this https://github.com/kubernetes-sigs/cloud-provider-kind/pull/227 , specially since is absorbing heavy weight dependencies

BenTheElder:
> I do not want to carry +4.5k files in the repo for a project like this https://github.com/kubernetes-sigs/cloud-provider-kind/pull/227 , specially since is absorbing heavy weight dependencies

I mean there is a substantial trade-off in not being able to actually review those, there are good reasons we vendor in k/k.

in k-sigs/kind I don't want to take the dependencies to begin with so tracking the changes is small (module proxy availability aside), but at least we split out the repos.

stmcginnis:
FWIW, I agree with @aojea on this. 

> And the module proxy doesn't guarantee availability of content

I see this as a good thing. If modules are yanked or somehow removed, it's good to know that and fix things.

aojea:
apologies for merging during the discussion, this project is very small and specific, so I took the liberty to merge, is not that we can do this in other situations or projects, but I have a PR https://github.com/kubernetes-sigs/cloud-provider-kind/pull/227 that I want to progress and was having some issues with having a large number of files

BenTheElder:
I'm still very -1, I think this is a huge risk that we are not reviewing the dependencies in this binary that runs as sudo and controls privileged containers.

in kind we can get away with it because I review the dependencies directly and they are small and few.

BenTheElder:
> I see this as a good thing. If modules are yanked or somehow removed, it's good to know that and fix things.

You would still detect this when the CI script to verify `go mod vendor`, fails, but the difference is you don't immediately lose the ability to build ...

BenTheElder:
> but I have a PR https://github.com/kubernetes-sigs/cloud-provider-kind/pull/227 that I want to progress and was having some issues with having a large number of files

there are better ways to handle that, as I mentioned in DM, to browse the files you inspect locally or commit by commit

github will not break with a large diff, you just don't look at that particular commit

we deal with this plenty of times in k/k with _far_ larger diffs

BenTheElder:
I also see it as a good thing that this is flagging that https://github.com/kubernetes-sigs/cloud-provider-kind/pull/227/files imports a ridiculous amount of code, which should not be necessary and does represent a supply-chain risk.

PR #250, opened 2025-06-05
support setting static LB IP via spec.LoadBalancerIP field

MatthiasWinzeler:
hi folks

first, many thanks for this great project - it's a great thing to be able to interact with real service `type: LoadBalancer` in local environments!

one thing that we'd love to see is the possibility to select static IPs for these services. currently, the IP for the services randomly picked from docker's `kind` network (usually `172.18.0.0/xx`). when docker is restarted, depending on the order of things in the `kind` network, a different IP is assigned. we have many other files that point to this IP that can't easily be changed (for example `kubeconfigs`), which then break.

would you folks be open to add the possibility to specify a static IP (ideally as part of an annotation, like many cloud providers do)? 

I created a small patch that does this:

https://github.com/MatthiasWinzeler/cloud-provider-kind

(run `go install github.com/MatthiasWinzeler/cloud-provider-kind@latest` to give it a spin)

use it like this:

```
apiVersion: v1
kind: Service
metadata:
  name: ingress-nginx-controller
  namespace: ingress-nginx
  labels:
    ..
  annotations:
    cloud-provider-kind.sigs.k8s.io/load-balancer-ip-address: 172.18.0.10
```

it probably only works for IPs in that Docker network (i.e. `172.18.0.x` in our case), but that seems to work and solves it for our use case.

Would you be willing to accept such an extension? If yes, I'm happy to brush up on the PR and add the necessary things - let me know.

### DIFF
--- a/.github/workflows/bats.yml
+++ b/.github/workflows/bats.yml
@@ -25,19 +25,31 @@ jobs:
       - name: Setup Bats and bats libs
         id: setup-bats
         uses: bats-core/bats-action@3.0.0
+
       - name: Bats tests
         shell: bash
         env:
          BATS_LIB_PATH: ${{ steps.setup-bats.outputs.lib-path }}
          TERM: xterm
-        run: bats -o _artifacts tests/
+        run: |
+          bats -o _artifacts tests/
+
+      - name: Bats tests (custom network)
+        shell: bash
+        env:
+          BATS_LIB_PATH: ${{ steps.setup-bats.outputs.lib-path }}
+          TERM: xterm
+        run: |
+          bats -o _artifacts-custom-network tests/custom-network/
 
       - name: Upload logs
         if: always()
         uses: actions/upload-artifact@v4
         with:
           name: kind-logs-${{ env.JOB_NAME }}-${{ github.run_id }}
-          path: ./_artifacts
+          path: |
+            ./_artifacts
+            ./_artifacts-custom-network
  
   bats_mac_tests:
     runs-on: macos-13
--- a/examples/loadbalancer_static_ip.yaml
+++ b/examples/loadbalancer_static_ip.yaml
@@ -1,0 +1,55 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: static-ip
+spec:
+  selector:
+    matchLabels:
+      app: static-ip
+  replicas: 1
+  strategy:
+    rollingUpdate:
+      maxSurge: 0
+      maxUnavailable: 1
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
+        app: static-ip
+    spec:
+      terminationGracePeriodSeconds: 30
+      affinity:
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            - labelSelector:
+                matchExpressions:
+                  - key: app
+                    operator: In
+                    values:
+                      - static-ip
+              topologyKey: kubernetes.io/hostname
+      containers:
+        - name: static-ip
+          image: registry.k8s.io/e2e-test-images/agnhost:2.39
+          args:
+            - netexec
+            - --http-port=80
+          ports:
+            - name: tcp
+              containerPort: 80
+              protocol: TCP
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: static-ip
+spec:
+  type: LoadBalancer
+  loadBalancerIP: REPLACE_WITH_STATIC_IP
+  selector:
+    app: static-ip
+  ports:
+    - name: tcp
+      port: 80
+      targetPort: 80
+      protocol: TCP
--- a/pkg/loadbalancer/server.go
+++ b/pkg/loadbalancer/server.go
@@ -239,6 +239,10 @@ func (s *Server) createLoadBalancer(clusterName string, service *v1.Service, ima
 	// Publish all ports in the host in random ports
 	args = append(args, "--publish-all")
 
+	if service.Spec.LoadBalancerIP != "" {
+		args = append(args, "--ip", service.Spec.LoadBalancerIP)
+	}
+
 	args = append(args, image)
 	// we need to override the default envoy configuration
 	// https://www.envoyproxy.io/docs/envoy/latest/start/quick-start/configuration-dynamic-filesystem
--- a/tests/README.md
+++ b/tests/README.md
@@ -5,4 +5,8 @@
 
 2. Install `kind` https://kind.sigs.k8s.io/
 
-3. Run `bats tests/`
+3. Run the tests:
+```
+bats tests/
+bats tests/custom-network
+```
--- a/tests/custom-network/setup_suite.bash
+++ b/tests/custom-network/setup_suite.bash
@@ -5,11 +5,17 @@ set -eu
 function setup_suite {
   export BATS_TEST_TIMEOUT=120
   # Define the name of the kind cluster
-  export CLUSTER_NAME="ccm-kind"
+  export CLUSTER_NAME="ccm-kind-custom-network"
 
-  export ARTIFACTS_DIR="$BATS_TEST_DIRNAME"/../_artifacts
+  export ARTIFACTS_DIR="$BATS_TEST_DIRNAME"/../../_artifacts-custom-network
   mkdir -p "$ARTIFACTS_DIR"
   rm -rf "$ARTIFACTS_DIR"/*
+
+  # create custom docker network
+  export KIND_EXPERIMENTAL_DOCKER_NETWORK=kind-static
+  docker network create \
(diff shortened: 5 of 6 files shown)


BenTheElder:
1.) Unreviewed APIs should be in x-k8s.io https://github.com/kubernetes/community/blob/e40e687af49428639351b45606f16c618c1e84fa/sig-architecture/api-review-process.md#voluntary
2.) I see the use case, but we've tried pretty hard to make sure this is 100% standards based ....
/assign @aojea

aojea:
@MatthiasWinzeler no need to use an annotation, this is already a standard field on the Service spec https://kubernetes.io/docs/concepts/services-networking/service/ (although the field is deprecated because it is ambiguous, but it does not mean it is not supported)

> Some cloud providers allow you to specify the loadBalancerIP. In those cases, the load-balancer is created with the user-specified loadBalancerIP. If the loadBalancerIP field is not specified, the load balancer is set up with an ephemeral IP address. If you specify a loadBalancerIP but your cloud provider does not support the feature, the loadbalancerIP field that you set is ignored.

So just parse the Service.spec and use that field instead of the annotation

aojea:
/ok-to-test

@MatthiasWinzeler seems you have to also sign the cla

MatthiasWinzeler:
hi @aojea 

thanks for getting back to me!

regarding using `service.Spec.LoadBalancerIP`: I first used that field, but the linter was pretty explicit that this field is deprecated:

```
        // Only applies to Service Type: LoadBalancer.
	// This feature depends on whether the underlying cloud-provider supports specifying
	// the loadBalancerIP when a load balancer is created.
	// This field will be ignored if the cloud-provider does not support the feature.
	// Deprecated: This field was under-specified and its meaning varies across implementations.
	// Using it is non-portable and it may not support dual-stack.
	// Users are encouraged to use implementation-specific annotations when available.
	// +optional
	LoadBalancerIP string `json:"loadBalancerIP,omitempty" protobuf:"bytes,8,opt,name=loadBalancerIP"`
```
https://github.com/kubernetes/kubernetes/blob/43bb11b5c1a998d110b391c028a01691e4e568b6/pkg/apis/core/types.go#L4840-L4849

I changed to using annotations because of the sentence `Users are encouraged to use implementation-specific annotations when available.`.

But for me it's fine to use either. What do you prefer? I'll change accordingly.

I also signed the CLA.

aojea:
> But for me it's fine to use either. What do you prefer? I'll change accordingly

yeah, heh, I was one of the persons on that discussion, but see my explanation here https://github.com/kubernetes/kubernetes/pull/107235#issuecomment-1320020675

The field from an API perspective was wrong because it leaked some details of some implementations and was not possible to implement dual stack with it, but since in cloud-provider-kind we are not going to expand the scope of the project for custom implementations with annotations, per example to assign static dual stack addresses to a loadbalancer (I also think there are no many use cases for that), I think that implementing this field is fair

MatthiasWinzeler:
@aojea makes sense - I changed it accordingly.

do you need something else - adding docs or a test?

btw, the tests appear to fail, but I think that might just be a flaky test that has nothing to do with this change:

```
I0608 09:04:27.468035 81886 networking_perf.go:260] Unexpected error: unable to parse iperf2 output from client pod iperf2-clients-4zfwj (node kind-worker): 
      <*errors.errorString | 0xc000d3d550>: 
      incorrect fields in the output: [time srcaddress srcport dstaddr dstport transferid istart iend bytes speed writecnt writeerr tcpretry tcpcwnd tcppcwnd tcprtt tcprttvar] (17 out of 15)
      {
          s: "incorrect fields in the output: [time srcaddress srcport dstaddr dstport transferid istart iend bytes speed writecnt writeerr tcpretry tcpcwnd tcppcwnd tcprtt tcprttvar] (17 out of 15)",
      }
```

aojea:
> I0608 09:04:27.468035 81886 networking_perf.go:260] Unexpected error: unable to parse iperf2 output from client pod iperf2-clients-4zfwj (node kind-worker): 

that error already has been fixed by https://github.com/kubernetes/kubernetes/pull/132180, it was indeed unrelated

> do you need something else - adding docs or a test?

see if you can add an e2e test here https://github.com/kubernetes-sigs/cloud-provider-kind/blob/main/tests/tests.bats, see the existing ones for examples, in this case it means you need to get the kind network address space and pick one ip from there, I think you can get that with some `docker inspect network --format...` and json templating

MatthiasWinzeler:
I added a simple e2e test. only for IPv6, should I add one for IPv6 as well? do we support that?

I think for the bats to rerun in e2e one of the maintainer needs to approve

aojea:
bats test is failing

```
> not ok 4 Static LoadBalancerIP
# (in test file tests/tests.bats, line 94)
#   `[ "$IP" = "$STATIC_IP" ]' failed
# Using static IP: 172.18.0.100
# deployment.apps/static-ip created
# service/static-ip created
# pod/static-ip-644dd77c89-2jx45 condition met
# Assigned IP:
```
no need to add another one for ipv6, I think this will get enough coverage for this feature and we can always revisit later

MatthiasWinzeler:
thanks @aojea for triggering again! but turns out, the failure only linux is more complex than thought:

on mac (where I was testing), setting the static IP works with default settings. on linux however, I get this error message:

```
docker: Error response from daemon: invalid config for network kind: invalid endpoint settings:
user specified IP address is supported only when connecting to networks with user configured subnets
```

So it seems that on Linux, we have to explicitly create a docker network, and passing it to both `kind` and `cloud-provider-kind` via the experimental `KIND_EXPERIMENTAL_DOCKER_NETWORK`.

so if I put all of this together, it works on linux:

```
 docker network create \
  --driver=bridge \
  --subnet=172.19.0.0/16 \
  kind-static

 cat <<EOF | KIND_EXPERIMENTAL_DOCKER_NETWORK=kind-static kind create cluster \
  --name ccm-kind           \
  -v7 --wait 1m --retain --config=-
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane
- role: worker
- role: worker
EOF

KIND_EXPERIMENTAL_DOCKER_NETWORK=kind-static cloud-provider-kind

bats --trace tests/tests.bats -f "Static LoadBalancerIP"
 ✓ Static LoadBalancerIP

1 test, 0 failures
```

so if we want to implement the feature on linux, I would suggest to:
1.) add a second set of e2e tests that create a docker network and run the new e2e test with static ips
2.) optionally: add a warning if `KIND_EXPERIMENTAL_DOCKER_NETWORK` is unset and a static IP is specified (but actually only necessary for linux?)

aojea:
Well, at least the output is the expected, if you set the field to a value and it does not work the lb fails, no?

We create a network but only add an ipv6 subnet https://github.com/kubernetes-sigs/kind/blob/3df64e784cc0ea74125b2a2e9877817418afa3af/pkg/cluster/internal/providers/docker/network.go#L149-L161 , the ipv4 seems to be autogenerated and per docker docs it seems is WAI

> 1.) add a second set of e2e tests that create a docker network and run the new e2e test with static ips

 maybe we can do some cluster per file instead one cluster per suite like I did today? I do not think we need to replicate the tests, this will be just for this case

> 2.) optionally: add a warning if KIND_EXPERIMENTAL_DOCKER_NETWORK is unset and a static IP is specified (but actually only necessary for linux?)

you need to do a docker inspect and validate that the subnet is an user configured subnet, the variable only indicates what network to use but we do not know if it is valid unless we check it, meanwhile setting a value that does not work fails it seems ok

MatthiasWinzeler:
@aojea 
> maybe we can do some cluster per file instead one cluster per suite like I did today? I do not think we need to replicate the tests, this will be just for this case

good idea, I added a second test suite for the ones that require custom networking. can you trigger the bats? it should work on linux now.

.github/workflows/bats.yml
```diff
@@ -30,14 +30,18 @@ jobs:
         env:
          BATS_LIB_PATH: ${{ steps.setup-bats.outputs.lib-path }}
          TERM: xterm
-        run: bats -o _artifacts tests/
+        run: |
+          bats -o _artifacts tests/
+          bats -o _artifacts-custom-network tests/custom-network/
```
aojea, line 35: split this into two steps, so it is clear when the job fails which one of the scenarios failed, the rest lgtm, so you can squash all commits too

MatthiasWinzeler: makes sense, done

stmcginnis:
Please squash commits - thanks!

MatthiasWinzeler:
@aojea thanks, addressed your commits & squashed, should be ready for another bats run

aojea:
/lgtm
/approve

We may add some section to explain how to use it @MatthiasWinzeler , up to you

PR #259, opened 2025-07-02
fix(tunnel): propagate half-closed TCP connections

justfalter:
This change ensures that tunneled TCP connections propagate shut downs when they occur, allowing the opposite party to be signaled appropriately.

A TCP connection can be "half-closed" when `A` shuts down `A->B` by sending a FIN to `B`, and `B` responds with an ACK. The connection becomes "fully-closed" when the same procedure is followed for the `A<-B` direction. When a TCP connection is half-closed in the `A->B` direction, data may continue to flow in the `A<-B` direction.

This signaling exists as a basic function of the TCP protocol, allowing either party to indicate that they are done sending data, and that their peer can finish up as they please. Data can continue to flow in the opposite direction until they are ready to shut down their side of the connection.

When cloud-provider-kind (`t`) acts as a tunnel for TCP connections between `A` and `B`, we have two connections `A-t` and `t-B`, and four flows:

- `A->t` and `t->B` (representing `A->B`).
- `A<-t` and `t<-B` (representing `A<-B`).

Prior to this change, when `A` shut down `A->t`, the tunnel would not shut down `t->B`, leaving `A-t` half-closed and `B-t` not closed in any way. The same is true for the opposite direction, as well.

After this change, when `A` shuts down `A->t`, cloud-provider-kind immediately shuts down `t->B`. The same is true when `B` shuts down `t<-B`, `A<-t` is immediately shut down. Flow directions are maintained independently from one-another.

Note: this change effectively restricts tunneled connections to only supporting `udp`, `tcp`, `tcp4`, and `tcp6` protocols. As far as I can tell, docker containers will not expose other protocol types (like unix domain sockets).

### DIFF
--- a/pkg/loadbalancer/tunnel.go
+++ b/pkg/loadbalancer/tunnel.go
@@ -120,7 +120,8 @@ func NewTunnel(localIP, localPort, protocol, remoteIP, remotePort string) *tunne
 
 func (t *tunnel) Start() error {
 	klog.Infof("Starting tunnel on %s %s", net.JoinHostPort(t.localIP, t.localPort), t.protocol)
-	if t.protocol == "udp" {
+	switch t.protocol {
+	case "udp":
 		localAddrStr := net.JoinHostPort(t.localIP, t.localPort)
 		udpAddr, err := net.ResolveUDPAddr("udp4", localAddrStr)
 		if err != nil {
@@ -136,8 +137,7 @@ func (t *tunnel) Start() error {
 				}
 			}
 		}()
-
-	} else if t.protocol == "tcp" || t.protocol == "tcp4" || t.protocol == "tcp6" {
+	case "tcp", "tcp4", "tcp6":
 		ln, err := net.Listen(t.protocol, net.JoinHostPort(t.localIP, t.localPort))
 		if err != nil {
 			return err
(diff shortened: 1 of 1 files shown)


aojea:
/ok-to-test

Thanks, just curious, how did you find out about this problem?

is envoy doing the right thing or do we also need to tune the envoy config?

justfalter:
> Thanks, just curious, how did you find out about this problem?

I have a grpc service hosted in a pod, and access it via cloud-provider-kind. I found that the client-side of a grpc server-stream was not getting an io.EOF when the service's pod was restarted. The client side of the server stream would remain connected to the tunnel without tearing down or erroring ... it'd just sit there, waiting for more server-provided messages to show up. 
 
> is envoy doing the right thing or do we also need to tune the envoy config?

I'm guessing that envoy is probably doing things properly as it has several orders of magnitude more users that cloud-provider-kind.

aojea:
> I'm guessing that envoy is probably doing things properly as it has several orders of magnitude more users that cloud-provider-kind.

I had an incidence some months ago with envoy because of problems with half closed connections, related to this https://github.com/istio/istio/issues/43297 , that is why I was asking if with current fix it works for you. We solved it setting a timeout for idle connections, but if this is working for you I prefer to keep the defaults

aojea:
/kind bug
/lgtm
/approve

Thanks

Issue #260, opened 2025-07-05
cloud-provider-kind handling IPs differently to kubelet in Docker


adrianmoisey:
When making a new Kind cluster, the defaults are to make a Docker network with both IPv4 and IPv6 enabled:

See:  https://github.com/kubernetes-sigs/kind/blob/a042b46587cdcf5121a18b5236a740c560e4e650/pkg/cluster/internal/providers/docker/network.go#L156-L157

Which gets executed from https://github.com/kubernetes-sigs/kind/blob/a042b46587cdcf5121a18b5236a740c560e4e650/pkg/cluster/internal/providers/docker/network.go#L68-L70

However, kubelet is configured to use IPv4 only with external cloud provider disabled.

When cloud-provider-kind is run, it detects the IPv4 and IPv6 address via Docker (see https://github.com/kubernetes-sigs/kind/blob/a042b46587cdcf5121a18b5236a740c560e4e650/pkg/cluster/internal/providers/docker/node.go#L55-L57), and updates the nodes with IPv6 addresses. 

This causes cloud-provider-kind and kubelet to keep changing the node's IP address, between IPv4 only and IPv4+IPv6, which can be seen by doing the following:
1. Make a new kind cluster via Docker
2. Run `cloud-provider-kind`
3. Run `kind get node kind-control-plane  -o jsonpath='{.status.addresses}{"\n"}' -w`

Notice an output similar to:
```
$ kubectl get node kind-control-plane  -o jsonpath='{.status.addresses}{"\n"}' -w
[{"address":"172.18.0.2","type":"InternalIP"},{"address":"kind-control-plane","type":"Hostname"}]
[{"address":"kind-control-plane","type":"Hostname"},{"address":"172.18.0.2","type":"InternalIP"},{"address":"fc00:f853:ccd:e793::2","type":"InternalIP"}]
[{"address":"172.18.0.2","type":"InternalIP"},{"address":"kind-control-plane","type":"Hostname"}]
[{"address":"kind-control-plane","type":"Hostname"},{"address":"172.18.0.2","type":"InternalIP"},{"address":"fc00:f853:ccd:e793::2","type":"InternalIP"}]
[{"address":"172.18.0.2","type":"InternalIP"},{"address":"kind-control-plane","type":"Hostname"}]
```

I'm happy to try fix this, but I'm not entirely sure what the fix is.

aojea:
[...]

PR #261, opened 2025-07-07
Only start CloudNodeController when nessesary

adrianmoisey:
Only start the CloudNodeController when the cloud provider taint exists on any node.

Fixes https://github.com/kubernetes-sigs/cloud-provider-kind/issues/260

I don't really know if this is the right solution. Happy to take suggestions.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -7,6 +7,7 @@ import (
 	"sync"
 	"time"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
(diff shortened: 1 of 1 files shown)


pkg/controller/controller.go
```diff
@@ -228,22 +231,42 @@ func startCloudControllerManager(ctx context.Context, clusterName string, kubeCl
...
-		cloud,
-		30*time.Second,
-		5, // workers
-	)
+	clustersInterface, ok := cloud.Clusters()
+	if !ok {
+		klog.Fatalf("Unable to use clusters interface: %v", err)
+	}
+
+	masterName, err := clustersInterface.Master(ctx, clusterName)
```
adrianmoisey, line 239: This seems overly complicated to get the masterName. I could change it to:
```
masterName := clusterName+'-control-plane'
```

Thoughts?

aojea, line 239: you need to identify the need of this controller by the Taint https://github.com/kubernetes/kubernetes/blob/3d13f39f24e5533286278f04702c69fbed448c35/staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go#L277C1-L278C1


just create a function that gets the clientset and list all the nodes and returns tur if any of the nodes has the cloud provider taint

aojea:
is this working fine now ? have yout tested it?

adrianmoisey:
> is this working fine now ? have yout tested it?

I've tested the "default" case (where the taint doesn't exist, and CloudNodeController is not started).
I haven't tested the case where the taint is present. I need to figure out how to get kind to do that.

aojea:
> I haven't tested the case where the taint is present. I need to figure out how to get kind to do that.

@adrianmoisey https://github.com/kubernetes-sigs/cloud-provider-kind/blob/main/kind.yaml

adrianmoisey:
Perfect, thanks! I'll give it a more thorough test

PR #295, opened 2025-10-02
Update gateway api to v1.4.0-rc.2

aojea:
This is a 3-steps process in cloud-provider kind

1. Update golang modules to use the latest clients: `go get -u sigs.k8s.io/gateway-api@v1.4.0-rc.2 && go mod tidy`
2. Update CRDs to install the latest APIs on the server:`hack/download-gateway-crds.sh v1.4.0-rc.2`
3. Update the github CI to use the latest conformance code v1.4.0-rc.2

Results can be seen in the corresponding github action, that builds conformance directly from the specified branch


stmcginnis:
Going with a release candidate probably isn't ideal, but it is better than where things stand today. And we should be able to move to an official, full release soon after. So looks good to me.

/lgtm

aojea:
They are going to cut the 1.4.0 final release today, @stmcginnis it will be nice if you can do the honor and land it here, you have the instructions in the description

stmcginnis:
> it will be nice if you can do the honor and land it here, you have the instructions in the description

Great, I will keep an eye out for it and update once it's available. :+1:

PR #299, opened 2025-10-06
Update gateway api to v1.4.0

stmcginnis:
Updates all go modules using `go get -u` and `go mod tidy`.

Updates to latest CRDs using the `hack/download-gateway-crds.sh` script

Updates GitHub workflow to use the v1.4.0 version.


aojea:
/lgtm

we should cat a new minor release

PR #305, opened 2025-10-24
gitignore DS_store files

BenTheElder:
https://en.wikipedia.org/wiki/.DS_Store

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,6 @@
 # vendor/
 bin/
 dist/
+
+# macOS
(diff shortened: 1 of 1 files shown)


stmcginnis:
I normally strongly recommend that everyone should have a [global gitignore](https://stackoverflow.com/a/7335487) to handle the common things to their development environment (`.DS_store`, `.vscode`, `.vim`, etc), but this seems to be a common enough surprise for people that it probably makes sense.

PR #312, opened 2025-10-27
GHA: Pin actions to specific SHA

stmcginnis:
Following security best practices, this pins the third party GitHub actions used to point to specific SHAs.


aojea:
hmm, the gateway class was not accepted ??

aojea:
it as a flake

Issue #313, opened 2025-10-28
Implement contextual logging


aojea:
Current logs shows the file name , and since there are multiple `controller.go` files it is not easy to understand which controller is doing what

```
I1028 08:17:12.051005  216621 controller.go:342] Updating HTTPRoute default/example-ingress-f2428e60d8
I1028 08:17:12.057950  216621 controller.go:342] Updating HTTPRoute default/example-ingress-45a335ded1
I1028 08:17:12.061574  216621 controller.go:342] Updating HTTPRoute default/example-ingress-f2428e60d8
I1028 08:17:12.091654  216621 gateway.go:50] Finished syncing gateway "default/kind-ingress-gateway" (45.651337ms)
I1028 08:17:12.091689  216621 gateway.go:73] Syncing Gateway default/kind-ingress-gateway, container kindccm-gw-b15a05d233de
I1028 08:17:12.134832  216621 gateway.go:50] Finished syncing gateway "default/kind-ingress-gateway" (43.141247ms)
I1028 08:18:12.046644  216621 controller.go:366] Started syncing gatewayclass "cloud-provider-kind" (92ns)
I1028 08:18:12.046707  216621 gateway.go:73] Syncing Gateway default/kind-ingress-gateway, container kindccm-gw-b15a05d233de
I1028 08:18:12.046731  216621 controller.go:342] Updating HTTPRoute default/example-ingress-f2428e60d8
I1028 08:18:12.050212  216621 controller.go:368] Finished syncing gatewayclass "cloud-provider-kind" (3.573521ms)
I1028 08:18:12.051513  216621 controller.go:342] Updating HTTPRoute default/example-ingress-45a335ded1
I1028 08:18:12.057963  216621 controller.go:342] Updating HTTPRoute default/example-ingress-45a335ded1
I1028 08:18:12.061462  216621 controller.go:342] Updating HTTPRoute default/example-ingress-f2428e60d8
I1028 08:18:12.090178  216621 ga
```

PR #315, opened 2025-11-02
Switch to contextual logging

stmcginnis:
This converts the cloud-provider-kind logging to follow the contextual logging pattern.

Closes: #313

### DIFF
--- a/cmd/app.go
+++ b/cmd/app.go
@@ -7,10 +7,10 @@ import (
 	"os"
 	"os/signal"
 	"runtime"
-	"strconv"
 	"strings"
 	"syscall"
 
+	"github.com/go-logr/logr"
 	"k8s.io/component-base/logs"
 	"k8s.io/klog/v2"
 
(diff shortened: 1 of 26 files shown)


aojea:
can we do first the context loggers without having to modify the functions signatures ... then in a follow up PR we can check how to structure the code to plumb down the loggers, but I want to avoid too much churn , specially about modifying functions

stmcginnis:
> can we do first the context loggers without having to modify the functions signatures

Not really cleanly. We would end up with a mix of contextual and non-contextual logging. There isn't really a way to get contextual logging without plumbing through.

If the original concern is just differentiating which controller the logging is coming from, we can just add the controller name to the logging in the respective `controller.go` files. Is that what you're looking for as far as immediate needs?

aojea:
> Is that what you're looking for as far as immediate needs?

yes, I should have been more precise sorry, just for identifying each controller ... in parallel, we should have a better architecture for logging, but I want you , @BenTheElder and I to agree first so we are consistent in the future

stmcginnis:
Slimmed down alternative in #323. Going to close this one for now.

PR #318, opened 2025-11-08
Move Gateway API section in README

stmcginnis:
The Gateway API support section was added inside the "How to use it" section of the README, disrupting the flow and making it look like things like "Running the provider" were subsections of Gateway API rather than initial setup instructions.

This moves the Gateway API support section just prior to "How to use it" to keep the flow of the setup instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,17 @@ Or using `compose.yaml` file:
 NET_MODE=kind docker compose up -d
 ```
 
+## Gateway API support (Alpha)
+
+This provider has alpha support for the [Gateway API](https://gateway-api.sigs.k8s.io/).
+It implements the `Gateway` and `HTTPRoute` functionalities.
+
+You can enable the Gateway API controllerby selecting the Gateway API release channel (standard/experimental):
+
+```sh
+cloud-provider-kind --gateway-channel standard
+```
+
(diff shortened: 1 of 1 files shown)


aojea:
Why the job fails?

>   I1108 13:38:49.482808 72206 wait.go:276] 	-> Node kind-worker2 [[[ Ready=true, Network(available)=false, Taints=[{node.cloudprovider.kubernetes.io/uninitialized true NoSchedule <nil>}], NonblockingTaints=node-role.kubernetes.io/control-plane ]]]
  I1108 13:38:49.482812 72206 wait.go:288] ==== node wait: 0 out of 3 nodes are ready, max notReady allowed 0.  Need 3 more before starting.
  I1108 13:39:19.481175 72206 e2e.go:211] Unexpected error: 
      <context.deadlineExceededError>: 
      context deadline exceeded
      
          {}
  [FAILED] in [SynchronizedBeforeSuite] - k8s.io/kubernetes/test/e2e/e2e.go:211 @ 11/08/25 13:39:19.481

stmcginnis:
Weird. That can't be coming from this change. We may have a flake there. I'll try to watch for that test and see if we get other odd failures.

/retest

Issue #327, opened 2025-11-20
Support custom envoyproxy image URL


jpolania:
Hello folks

We have tested and so far is working; but you have hardcoded proxy docker url here
https://github.com/kubernetes-sigs/cloud-provider-kind/blob/main/pkg/images/images.go#L4

This new feature will add a new parameter as `--proxy-image` where you can overwrrite the default ProxyImage = docker.io/envoyproxy/envoy:v1.33.2

This is handy if you want to use another version of envoy or your Docker Registry is on a corporate network

jpolania:
I have created this PR for review
https://github.com/kubernetes-sigs/cloud-provider-kind/pull/328

aojea:
we decided to only allow to override the registry, unfortunately this project grew so holistically that there is much coupling with the image details 

Thanks for your contribution and for your help with the original implementation

jpolania:
> we decided to only allow to override the registry, unfortunately this project grew so holistically that there is much coupling with the image details
> 
> Thanks for your contribution and for your help with the original implementation

Understood so thank you !!!

PR #328, opened 2025-11-20
feat: add custom proxy image url

jpolania:
This small PR allows to you to use a custom `proxy-image` value to be parametrize when is launched as 

`sudo nohup ./cloud-provider-kind --proxy-image docker.io/envoyproxy/envoy:v1.36.2 --v=4 > /tmp/cloud-provider-logs.txt 2>&1 &`

logs from /tmp/cloud-provider-logs.txt

[cloud-provider-logs.txt](https://github.com/user-attachments/files/23647493/cloud-provider-logs.txt)

Fixes: https://github.com/kubernetes-sigs/cloud-provider-kind/issues/327

### DIFF
--- a/cmd/app.go
+++ b/cmd/app.go
@@ -55,11 +55,6 @@ func init() {
 }
 
 func Main() {
-	// Parse command line flags and arguments
-	flag.Parse()
-	flag.VisitAll(func(flag *flag.Flag) {
-		klog.Infof("FLAG: --%s=%q", flag.Name, flag.Value)
-	})
 
 	// Validate Proxy Image
 	if proxyImage == "" {
@@ -76,6 +71,12 @@ func Main() {
 		default:
 		}
 	}
+
+	// Parse command line flags and arguments
+	flag.Parse()
+	flag.VisitAll(func(flag *flag.Flag) {
+		klog.Infof("FLAG: --%s=%q", flag.Name, flag.Value)
+	})
 
 	// don't allow subcommands after flags
 	if len(flag.Args()) > 0 {


cmd/app.go
```diff
@@ -53,22 +55,25 @@ func init() {
...
+	})
+
+	config.DefaultConfig.ProxyImage = proxyImage
+	images.Images["proxy"] = proxyImage
+
 	// Parse subcommands if exist
-	if len(os.Args) > 1 {
-		switch os.Args[1] {
+	if len(flag.Args()) > 0 {
+		switch flag.Args()[0] {
```
aojea: why this change?

aojea:
/ok-to-test

lgtm but it is better to adapt the code to avoid having multiple global variables

cmd/app.go
```diff
@@ -41,6 +42,7 @@ func init() {
 	flag.BoolVar(&enableLBPortMapping, "enable-lb-port-mapping", false, "enable port-mapping on the load balancer ports")
 	flag.StringVar(&gatewayChannel, "gateway-channel", "standard", "define the gateway API release channel to be used (standard, experimental, disabled), by default is standard")
 	flag.BoolVar(&enableDefaultIngress, "enable-default-ingress", true, "enable default ingress for the cloud provider kind ingress")
+	flag.StringVar(&proxyImage, "proxy-image", "docker.io/envoyproxy/envoy:v1.33.2", "proxy image to use for load balancers")
```
aojea: lets' make this `docker.io/envoyproxy/envoy:v1.33.2` a constant `defaultProxyImage` or similar

BenTheElder:
Is this intended to support pulling from a mirror, or to use different versions of envoy / modified images?

We should probably be clear about what is supported.

aojea:
> Is this intended to support pulling from a mirror, or to use different versions of envoy / modified images?
> 
> We should probably be clear about what is supported.

@BenTheElder what is your concern? 

I think that some people on airgap will use the mirror thing.
I think that some people developing with envoy may want to test new versions? I'm unclear on this one, but since we added gateway I can see this can be a thing

jpolania:
> Is this intended to support pulling from a mirror, or to use different versions of envoy / modified images?
> 
> We should probably be clear about what is supported.

My intention is pulling from a `mirror docker registry` which is behind of a company network I do consulting; also I realized it can be used to use a different version of envoy as well.

jpolania:
> > Is this intended to support pulling from a mirror, or to use different versions of envoy / modified images?
> > We should probably be clear about what is supported.
> 
> @BenTheElder what is your concern?
> 
> I think that some people on airgap will use the mirror thing. I think that some people developing with envoy may want to test new versions? I'm unclear on this one, but since we added gateway I can see this can be a thing

All this change is to parametrize `image-proxy` as the docker registry can be different but agreed this open the case for more parametrization in the future

jpolania:
@aojea all your suggestions (except the first one) has been actioned on my 2nd commit
https://github.com/kubernetes-sigs/cloud-provider-kind/pull/328/commits/7a33e87d0f8084a66857f6420eb66b27d2b57ce6

BenTheElder:
> I think that some people developing with envoy may want to test new versions? I'm unclear on this one, but since we added gateway I can see this can be a thing

Yeah, for this I think we might want to at least document that it's primarily intended for mirroring and we don't make any compatibility claims with different envoy versions / have no intention to test and support these.

I don't think we should block people from using arbitrary images, but we should avoid creating the perception that we actively *support* using customized envoy images, and be clear that envoy itself is an implementation detail.

> All this change is to parametrize image-proxy as the docker registry can be different but agreed this open the case for more parametrization in the future

In this case, it might make more sense to offer an option that generically swaps the host for a registry mirror for all images, which also avoids encoding the coupling to the envoy image specifically ...

This is what we do for e2e.test and airgapping in kubernetes/kubernetes, we allow overriding the host *for all images* but we don't allow overriding the image tags etc

BenTheElder:
If we make the ability to use custom envoy versions, and we don't say anything about the expected support, the default assumption from many users will be that whatever is possible is supported, and then it will be painful in the future when we break them (e.g. we may need to rebase on another loadbalancer someday)

jpolania:
> If we make the ability to use custom envoy versions, and we don't say anything about the expected support, the default assumption from many users will be that whatever is possible is supported, and then it will be painful in the future when we break them (e.g. we may need to rebase on another loadbalancer someday)

Eventually you need to open up for parametrization, so what I have done is updating README.md to add that warning like

> [!WARNING]
> Using a custom mirror URL and/or different Envoy version is not supported by `cloud-provider-kind` and should be used at your own risk.

hope that satisfy your concern

jpolania:
@BenTheElder  and @aojea 
Thanks for your feedback I completed the changes as requested. Personally I don't want to spent too much time on this PR so this is a small contribution out of good will.

If I noticed too much fuzz about; then I'll stop contribute, move my fork private and enventually will deploy this internally in the organization I have to do this change.

Let's keep the common sense above all

Thanks

aojea:
@jpolania sorry for caught you in the middle of the discussion, and thank for the contribution, if something else need to be modified I can take over, and I will keep your signature, do not worry about it, is just that things like this can always bite us later

> If we make the ability to use custom envoy versions, and we don't say anything about the expected support, the default assumption from many users will be that whatever is possible is supported, and then it will be painful in the future when we break them (e.g. we may need to rebase on another loadbalancer someday)

@BenTheElder I see this as `kind create cluster --image` , and I also find it useful as I had issues with some envoy versions that had a bug that crashes with udp. I'm not much worried about people using custom images and not working if is well documented as it is in this PR

I just have one comment with the flags / subcomands but it LGTM

jpolania:
> @jpolania sorry for caught you in the middle of the discussion, and thank for the contribution, if something else need to be modified I can take over, and I will keep your signature, do not worry about it, is just that things like this can always bite us later
> 
> > If we make the ability to use custom envoy versions, and we don't say anything about the expected support, the default assumption from many users will be that whatever is possible is supported, and then it will be painful in the future when we break them (e.g. we may need to rebase on another loadbalancer someday)
> 
> @BenTheElder I see this as `kind create cluster --image` , and I also find it useful as I had issues with some envoy versions that had a bug that crashes with udp. I'm not much worried about people using custom images and not working if is well documented as it is in this PR
> 
> I just have one comment with the flags / subcomands but it LGTM

Thank you @aojea feel free to take over, add the disclaimers, requirements, testing, etc to satisfy your community needs

I can say the functionality has been enabled.  Today I merged my forked repo to main (on another feature branch) and able to compile the binary and effectively the `--proxy-image $MIRROR_REGISTRY_URL/envoyproxy/envoy:v1.33.2` is working for my client.

Actioned your request
"I just have one comment with the flags / subcomands"

Thanks all and keep up the good work

aojea:
superseded by https://github.com/kubernetes-sigs/cloud-provider-kind/pull/330

I make sure @jpolania your commits author was included, sorry for the change of the approach , but after talking with @BenTheElder making it a flag. option will bit us in the future, and break users that may define their scripts with that flag.
The env variable is widely used in kubernetes and I think from the end users perspective does not change much

PR #330, opened 2025-11-24
Add option to use a mirror registry

aojea:
Replaces https://github.com/kubernetes-sigs/cloud-provider-kind/pull/328

Cloud provider kind makes a lot of assumptions on the image, injects binaries or depends on envoy paths, and we also have changed the dataplane before (haproxy -> envoy).

Using flags also creates a tightly coupling and every change there will be a breaking change, since the cli is our main API with end users.

This uses an environment variable to be able to override the registry used for the envoy image.
It also refactors the code to depend on the singleton global variable for the configuration and use cobra to handle the subcomannds and flags

/cc @jpolania
/assign @BenTheElder

### DIFF
--- a/cmd/app.go
+++ b/cmd/app.go
@@ -2,7 +2,6 @@ package cmd
 
 import (
 	"context"
-	"flag"
 	"fmt"
 	"os"
 	"os/signal"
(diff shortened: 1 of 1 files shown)


aojea:
now is ok, it had a panic trying to merge the klog flags with the cobra ones

stmcginnis:
Great! I should have looked deeper at the failures, but assumed since everything failed it was an infra glitch. Makes sense.

/lgtm

aojea:
/hold cancel

let's get this in

PR #340, opened 2025-12-08
Correct filepath/path for embed.FS on Windows

evankanderson:
Surprisingly, even on Windows, `embed.FS` uses unix-style paths.  This switches the embedded path logic for finding CRDs to work on Windows.  Without this, you get the following error:

```
E1208 13:53:59.568349    7968 controller.go:294] "Failed to install Gateway API CRDs" err="error processing embedded CRDs from crds\\\\standard: error walking embedded fs at \\"crds\\\\\\\\standard\\": open crds\\\\standard: file does not exist" cluster="kind"
```

Which, in turn, prevents the cloud-provider-kind from starting up and assigning IPs for LoadBalancer services.

### DIFF
--- a/pkg/gateway/crd_manager.go
+++ b/pkg/gateway/crd_manager.go
@@ -6,7 +6,7 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
-	"path/filepath"
+	"path"
 	"strings"
 
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


aojea:
/ok-to-test

I can not test it so I assume you already did it, right?

evankanderson:
Yes. I'm currently running this (as administrator).

I did find one other challenge with my setup where I had a node with a hostport that conflicted with the gateway, but that was not `cloud-provider-kind`s fault.

stmcginnis:
I don't think any of us have the environment to test this ourselves, but based on your own testing, and that the change looks correct, this should be good.

/lgtm
/approve

PR #346, opened 2025-12-22
Bump go, k8s, and kind versions

stmcginnis:
Go version bump needed in order to pick up k8s.io library updates that have now moved ahead to 1.25 (#345).


aojea:
here we go

/lgtm
/approve

PR #348, opened 2025-12-26
Feature Implement Unix Domain Socket based connectivity for Envoy->XDS server for gateway.

### LINKED ISSUE #309: Improve XDS Server connectivity
Envoys for Gateway uses the XDS protocol and cloud-provider-kinds acts as an XDS server.

Currently cloud provider kind tries to figure out its IP address and Port

https://github.com/kubernetes-sigs/cloud-provider-kind/blob/37f4b10583b9e17331370197c9b943882147af9a/pkg/gateway/controller.go#L671-L717

and configures envoy to use them, together with the well known lima and docker ips to reach the host, in case of platforms like Windows or Mac where containers run inside a VM and cloud provider kind in the host

https://github.com/kubernetes-sigs/cloud-provider-kind/blob/37f4b10583b9e17331370197c9b943882147af9a/pkg/gateway/envoy.go#L32-L83

Ideally, to be more resilient, we can export an additional Port in the envoy container, and be able to create a reverse tunnel connecting to that exposed port, so we do not need to worry about the dynamic XDS server IP:Port, it will be fixed.

It seems that envoy has recently added some reverse tunnel functionality https://github.com/envoyproxy/envoy/issues/33320 so it will be good to understand if that can be used and implement it, and drop all the existing brittle heuristics

ashishkpathak-connect:
Implement Unix Domain Socket based connectivity for Envoy to connect to XDS server for gateway.

Fixes #309

### DIFF
--- a/pkg/gateway/controller.go
+++ b/pkg/gateway/controller.go
@@ -20,7 +20,7 @@ import (
 	"context"
 	"fmt"
 	"net"
-	"sort"
+	"os"
 	"sync/atomic"
 	"time"
 
@@ -67,6 +67,7 @@ import (
 const (
 	controllerName = "kind.sigs.k8s.io/gateway-controller"
 	GWClassName    = "cloud-provider-kind"
+	GWSocketName   = "/var/run/cloudproviderkind.sock"
 	maxRetries     = 5
 	workers        = 5
 )
@@ -120,11 +121,9 @@ type Controller struct {
 	referenceGrantLister       gatewaylistersv1beta1.ReferenceGrantLister
 	referenceGrantListerSynced cache.InformerSynced
 
-	xdscache        cachev3.SnapshotCache
-	xdsserver       serverv3.Server
-	xdsLocalAddress string
-	xdsLocalPort    int
-	xdsVersion      atomic.Uint64
+	xdscache   cachev3.SnapshotCache
+	xdsserver  serverv3.Server
+	xdsVersion atomic.Uint64
 
 	tunnelManager *tunnels.TunnelManager
 }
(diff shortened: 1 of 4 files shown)


ashishkpathak-connect:
**Tests**

Test - 
[cpk-uds-test.pdf](https://github.com/user-attachments/files/24348974/cpk-uds-test.pdf)


Ref - https://github.com/kubernetes-sigs/cloud-provider-kind/issues/309#issuecomment-3693090592

aojea:
how will this work on windows or mac where the containers run in a vm ?

ashishkpathak-connect:
> how will this work on windows or mac where the containers run in a vm ?

@aojea The test report(pdf) I've attached is from Mac itself. I am unsure how I can test for windows currently. Whats your view on this? It wouldn't work on windows? Can we somehow try testing this PR for windows?

aojea:
here is the thing, cloud provider kind can run standalone in the host or as a container.
If you run the cloud provider kind in the mac, then somehow you need to mount those file descriptors from the containers vm into the mac host ... are you testing running cloud provider kind as a docker container in mac?

ashishkpathak-connect:
> here is the thing, cloud provider kind can run standalone in the host or as a container. If you run the cloud provider kind in the mac, then somehow you need to mount those file descriptors from the containers vm into the mac host ... are you testing running cloud provider kind as a docker container in mac?

Thanks @aojea I was testing by running cloud-provider-kind on the Mac host by directly running the binary with root perms. I get your point now, if we run cpk as docker container and if we use Unix Domain Socket, then we need somehow to mount the socket to cpu container. 
Apologies, I didnt think of the scenario when cpk is run as container. I always thought its invoked as a binary and designed/tested above based on that.

aojea:
> Apologies, I didnt think of the scenario when cpk is run as container. I always thought its invoked as a binary and designed/tested above based on that.

no worries, the project grew in scope based on user feedback and this seems to be a common case now, is normal most people skipped it, but I love this solution based on sharing the unix domain socket, wonder if we can generalize it 🤔

ashishkpathak-connect:
> > Apologies, I didnt think of the scenario when cpk is run as container. I always thought its invoked as a binary and designed/tested above based on that.
> 
> no worries, the project grew in scope based on user feedback and this seems to be a common case now, is normal most people skipped it, but I love this solution based on sharing the unix domain socket, wonder if we can generalize it 🤔

With Generalize, were you referring to be able to use this solution across windows, Mac, linux + host/container right? Do you have any ideas?

aojea:
> With Generalize, were you referring to be able to use this solution across windows, Mac, linux + host/container right? Do you have any ideas?

that is the tricky problem, people expect the project to work in any environment, so we can not develop a feature that only works in some of them or we break the others ... the reverse tunnel should work because we can expose ports from the container, so we are able to connect to specific ports in the containers from the host

ashishkpathak-connect:
> > With Generalize, were you referring to be able to use this solution across windows, Mac, linux + host/container right? Do you have any ideas?
> 
> that is the tricky problem, people expect the project to work in any environment, so we can not develop a feature that only works in some of them or we break the others ... the reverse tunnel should work because we can expose ports from the container, so we are able to connect to specific ports in the containers from the host

Totally agreed @aojea 
Reviewing the doc https://www.envoyproxy.io/docs/envoy/latest/configuration/other_features/reverse_tunnel carefully,
it doesn't seem to provide ReverseTunnel connectivity for xDS. Reverse Tunnel is for data connectivity b/w two envoy and not for xDS protocol.

In our case, we have setup like below
Envoy ---(XDS)--- Gateway server

As per my understanding for our setup, wherein we want to expose a port of Envoy and GatewayServer to make a gRPC dialup to Envoy's port which then Envoy uses it as a reverse connection for xDS is not possible currently.

aojea:
> it doesn't seem to provide ReverseTunnel connectivity for xDS.

well, these are network tunnels and I was wondering if envoy dataplane is able to reverse tunnel any kind of traffic, I was assuming envoy will not have a real notion if it is its own traffic or not ... then it may work

PR #351, opened 2026-01-05
Update README for Gateway API support clarity

aojea:
Clarified Gateway API support details and usage instructions.

Gateway API is enabled by default and passes all conformance tests, it is a gateway implementation that works on kind clusters, so that is the context of this functionality


rikatz:
/lgtm
Thanks Antonio!

Issue #355, opened 2026-01-09
Ingress supports WebSockets


singhwang:
In the latest version, I can perfectly access the HTTP service using Ingress, but not with WebSocket; the current implementation cannot connect to the WebSocket service.

aojea:
@singhwang can you please provide a reproducer and more detail?

singhwang:
> [@singhwang](https://github.com/singhwang) can you please provide a reproducer and more detail?

This issue can be reproduced by deploying the open-source version of Rancher on a kind Kubernetes cluster. Rancher itself is a containerized application that supports deployment on Kubernetes using Helm, and its relatively comprehensive functionality makes it suitable for reproducing this issue. Opening the container terminal and viewing container logs in the Rancher UI are both implemented via the WebSocket interface. If using the corresponding Ingress implementation for Treafik or Nginx, the WebSocket interface works correctly. However, after switching to cloud-provider-kind, all HTTP interfaces work normally, but all WebSocket interfaces fail.

**Special note: By default, a Kubernetes cluster that installs Rancher Helm Chart must have LoadBalancer and Ingress capabilities.**

https://ranchermanager.docs.rancher.com/getting-started/quick-start-guides/deploy-rancher-manager/helm-cli#install-rancher-with-helm

aojea:
how do you know is not a rancher problem? rancher inside kind does not look like a simple integration

singhwang:
> how do you know is not a rancher problem? rancher inside kind does not look like a simple integration

I also use kind, but instead of cloud-provider-kind, I use it with metallb and ingress (treafik and nginx). The Rancher deployment I mentioned works fine. After switching to cloud-provider-kind, LB and ingress are mostly normal, only the websocket doesn't work.

I'm not suggesting that Rancher integration be supported, but Rancher itself has ready-made application functions that can reproduce this issue. If you have other, simpler WebSocket applications, you can try them.I believe the issue will still exist.

aojea:
@singhwang can you please test this patch and validate if this solves the issue https://github.com/kubernetes-sigs/cloud-provider-kind/pull/356 ?

singhwang:
> can you please test this patch and validate if this solves the issue [#356](https://github.com/kubernetes-sigs/cloud-provider-kind/pull/356) ?

I have tested this patch by deploy rancher and can confirm that it resolves issue [#356](https://github.com/kubernetes-sigs/cloud-provider-kind/pull/356) .

[...]

PR #356, opened 2026-01-12
support websocket upgrade

aojea:
Support websocket upgrades on http connections

Fixes: #355

/hold

/cc @singhwang


aojea:
/hold cancel

tested by @singhwang

Issue #359, opened 2026-01-21
Envoy Panic threshold conflicting with externalTrafficPolicy: Local


adberger:
Hi there

When having multiple kind (worker) nodes (e.g. 3) and `externalTrafficPolicy` set to `Local`, traffic gets routed to unhealthy nodes.

kind cluster:
```yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
name: nginx-cluster
nodes:
- role: control-plane
- role: worker
- role: worker
- role: worker
networking:
  # ipFamily: dual
  disableDefaultCNI: true
  kubeProxyMode: "none"
```

nginx gateway fabric:
```
helm upgrade --install -n nginx-gateway --create-namespace --version 2.3.0 -f nginx-gateway/values.yaml ngf oci://ghcr.io/nginx/charts/nginx-gateway-fabric
```

nginx-gateway/values.yaml:
```yaml
nginx:
  replicas: 1
  service:
    externalTrafficPolicy: Local
```

gateway:
```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  name: ngf-gateway
  namespace: default
spec:
  gatewayClassName: nginx
  listeners:
  - protocol: HTTP
    port: 80
    name: ngf-http
    allowedRoutes:
      namespaces:
        from: Same
```

podinfo:
```
helm upgrade --install -n default --version 6.9.4 -f podinfo/values.yaml podinfo podinfo/podinfo
```

podinfo/values.yaml:
```yaml
httpRoute:
  enabled: true
  hostnames:
  - example.local
  parentRefs:
  - name: ngf-gateway
    namespace: default
    sectionName: ngf-http
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: /
```

In this setup calling the HTTPRoute several times, fails quite often:
```
export LB_IP=$(kubectl get svc -n default -l gateway.networking.k8s.io/gateway-name=ngf-gateway -o jsonpath='{.items[0].status.loadBalancer.ingress[0].ip}')
for _ in {1..10}; do
  curl -s -o /dev/null -w "%{http_code}\n" --header 'Host: example.local' http://${LB_IP}
  sleep 0.5
done
```

result:
```
000 # fails with (curl: (52) Empty reply from server)
000 # fails with (curl: (52) Empty reply from server)
000 # fails with (curl: (52) Empty reply from server)
[...]

PR #360, opened 2026-01-22
fix: disable Envoy healthy panic threshold

aojea:
By default, Envoy enables a "healthy panic threshold" of 50%. This causes Envoy to enter "panic mode" and route traffic to all upstream hosts—including unhealthy ones—if the percentage of healthy hosts drops below this threshold.

In a Kubernetes environment, this behavior defeats the purpose of Readiness probes and the Service controller. Kubernetes explicitly manages endpoint health; if a Pod is not ready, it is removed from the EndpointSlice/Endpoints API. The Cloud Provider LoadBalancer is expected to strictly respect this state.

This commit sets the `healthy_panic_threshold` to 0% for both Gateway API (xDS) and LoadBalancer (static) configurations. This ensures the LoadBalancer behaves deterministically and "fails closed" (returns 503s) rather than "failing open" (sending traffic to broken Pods) when the cluster is in a degraded state.

Fixes: https://github.com/kubernetes-sigs/cloud-provider-kind/issues/359

### DIFF
--- a/pkg/gateway/gateway.go
+++ b/pkg/gateway/gateway.go
@@ -16,6 +16,7 @@ import (
 	endpointv3 "github.com/envoyproxy/go-control-plane/envoy/config/endpoint/v3"
 	listenerv3 "github.com/envoyproxy/go-control-plane/envoy/config/listener/v3"
 	routev3 "github.com/envoyproxy/go-control-plane/envoy/config/route/v3"
+	typev3 "github.com/envoyproxy/go-control-plane/envoy/type/v3"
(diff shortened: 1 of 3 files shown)


aojea:
/assign @adberger

Please test it and LGTM once you verifies it works as expected, I tested locally and it works fine for me

stmcginnis:
Looks good to me, and the logic makes sense. Would need to hear from @adberger though.

/hold
/lgtm

adberger:
@aojea I've successfully tested the changes in this PR.
Thank you very much!

stmcginnis:
Thanks!

/unhold

PR #373, opened 2026-02-20
Fix Gateway CRD manager choking on kustomization.yaml

lizbound:
When using the Experimental gateway channel, the directory contains a `kustomization.yaml` that the crd manager attempts to load as a k8s object, which fails. 

This PR adds a one-off exception for that file (the only one present in [the directory](https://github.com/kubernetes-sigs/cloud-provider-kind/tree/e6c33ec3e6867a491a3dfee4a55ff293e54beba0/pkg/gateway/crds/experimental)) and adds tests for good measure.

### DIFF
--- a/pkg/gateway/crd_manager.go
+++ b/pkg/gateway/crd_manager.go
@@ -79,6 +79,11 @@ func (m *CRDManager) InstallCRDs(ctx context.Context, channelDir config.GatewayR
 			return nil // Continue walking
 		}
 
+		// Skip the kustomize files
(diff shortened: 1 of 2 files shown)


aojea:
/ok-to-test

Thanks, you need to sign the CLA

An alternative is just to avoid copying it https://github.com/kubernetes-sigs/cloud-provider-kind/blob/main/hack/download-gateway-crds.sh but that seems more complicated than this

lizbound:
Working on the CLA, just joined a new company and didn't realise there wasn't one on file

Issue #404, opened 2026-05-04
cloud-provider-kind does not respect KIND_EXPERIMENTAL_PROVIDER=podman if Docker is installed


avosepp:
Hello All, 

Trying to run a Kind cluster with Podman on the latest MacOS today and came across this. 

```zsh
E0504 12:22:31.375083   26245 controller.go:71] "Error listing clusters, retrying..." err="failed to list clusters: command \"docker ps -a --filter label=io.x-k8s.kind.cluster --format '{{.Label \"io.x-k8s.kind.cluster\"}}'\" failed with error: exit status 1"
```

This failure occurs when trying to run `cloud-provider-kind`. I have `alias docker=podman` and the `KIND_EXPERIMENTAL_PROVIDER=podman` env var in both my shell and the `sudo` shell. `cloud-provider-kind` was still attempting to use Docker and erroring with the above error. 

I ran across #242 where @colinjlacy pointed out that completely uninstalling Docker from their machine resolved the issue. I tested the same on my machine and was able to verify. After Docker was removed, my machine began to respect the `KIND_EXPERIMENTAL_PROVIDER=podman` env var. 

It seems like in the code somewhere the check of Docker being installed is overriding the env var or even the alias. 

I think it's reasonable for a user to have both Docker and Podman installed on their machine. It's a little difficult to troubleshoot.

stmcginnis:
Are you running with `sudo`? The question of how `KIND_EXPERIMENTAL_PROVIDER` was being set was never answered in #242.

avosepp:
`sudo KIND_EXPERIMENTAL_PROVIDER=podman cloud-provider-kind`

This did not work for me. Would you like me to try some alternatives and report back?

EDIT: Did not work for me when Docker was installed. It worked after Docker was uninstalled.

stmcginnis:
Let's try this:

```
export KIND_EXPERIMENTAL_PROVIDER=podman
sudo -E cloud-provider-kind
```

avosepp:
```zsh
export KIND_EXPERIMENTAL_PROVIDER=podman
```

❌ With Docker Installed

```zsh
sudo -E cloud-provider-kind
Enter PIN for 'Certificate For PIV Authentication (Yubico PIV Authentication)':
I0504 13:27:46.955016   54670 app.go:77] FLAG: --enable-default-ingress="true"
[...]

PR #406, opened 2026-05-04
Explicitly set provider runtime

stmcginnis:
Improves logic around which runtime to use, making sure to always account for KIND_EXPERIMENTAL_PROVIDER environment variable setting.

Related to: #404

### DIFF
--- a/cmd/app.go
+++ b/cmd/app.go
@@ -165,14 +165,38 @@ func runE(cmd *cobra.Command, args []string) error {
 	config.DefaultConfig.ControlPlaneConnectivity = config.Portmap
 
 	// initialize kind provider
+	// Check KIND_EXPERIMENTAL_PROVIDER directly, as the kind CLI does, so it
+	// takes effect regardless of auto-detection results.
 	var option cluster.ProviderOption
-	switch p := container.Runtime(); p {
-	case "podman":
-		option = cluster.ProviderWithPodman()
-	case "nerdctl", "finch", "nerdctl.lima":
-		option = cluster.ProviderWithNerdctl(p)
-	default:
-		option = cluster.ProviderWithDocker()
+	if p := os.Getenv("KIND_EXPERIMENTAL_PROVIDER"); p != "" {
+		switch p {
+		case "podman":
+			klog.Infof("Using podman provider due to KIND_EXPERIMENTAL_PROVIDER")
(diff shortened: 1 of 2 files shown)


aojea:
@avosepp are you able to check this patch to confirm it fixes the problem?

avosepp:
[log.txt](https://github.com/user-attachments/files/27375965/log.txt)

I compiled this with `make build` and ran it with `sudo ./cloud-provider-kind --enable-lb-port-mapping` and got this output. It seems like it is throwing a new error, but it did correctly attempt to use podman. 

I'm not sure if this is a related to the way I'm running the app or not. But maybe these errors will help you.

stmcginnis:
Thanks - that error is something unrelated to proper runtime detection, so it looks like this does do what it's intended to do.

stmcginnis:
Guessing the error is maybe an environment issue? Have you seen this before @aojea?

```
"Failed to watch" err="failed to list *v1.ReferenceGrant: the server could not find the requested resource (get referencegrants.gateway.networking.k8s.io)" logger="UnhandledError" reflector="pkg/mod/k8s.io/client-go@v0.36.0/tools/cache/reflector.go:343" type="*v1.ReferenceGrant"
```

Well past the provider selection though, so this is definitely something unrelated.

avosepp:
Wait! I'm sorry. I forgot I had Docker uninstalled. ~When reinstalling it I got the Docker ps error again. My brain jumped to testing the built binary and forgot to adjust my system.~

Ignore this. I uninstalled docker, then ran with sudo and the ENV flag in the same line. And it did correctly recognize it need to run with podman. Then I had the same errors in the log.txt I posted above. 

So it looks like yes. This fix works. Quite possibly the errors thrown in the log I uploaded above are just due to the way I built or ran the software when I self-compiled it.

aojea:
@stmcginnis I think that right now we can remove init() entirely.
Then we move the autodetection logic from init() into an explicit helper, e.g. container.DetectRuntime()
and in cmd/app.go, do:
1. if KIND_EXPERIMENTAL_PROVIDER is set, use it
2. else p := container.DetectRuntime()
3. build the kind provider option from p
4. call container.SetRuntime(p)

aojea:
/lgtm
/approve

Thanks

PR #372, opened 2026-02-16
update deps

aojea:
Unit test fails if you do not update the generated code to the latest , because of the new feature in client go, see https://github.com/kubernetes-sigs/cloud-provider-kind/pull/345#issuecomment-3907664068

PR #411, opened 2026-05-12
Update Gateway CRDs to v1.5.0

lelandbatey:
Updates the Gateway API CRDs from version v1.4.1 to v1.5.0 using:

```
GIT_VERSION=v1.5.0 ./hack/download-gateway-crds.sh
```

# Notes for reviewers

It's unclear what the policy is around when cloud-provider-kind upgrades its Gateway CRDs as new versions of Kubernetes Gateway API come out. The only info I found on why past upgrades of these CRDs have happened seems to be in the prior PR #372 which seems to have been motivated by "unit tests fail [due to an upstream change] if you do not update the generated code to the latest".

This PR is not motivated by broken tests, it's motivated by a desire for `cloud-provider-kind` (hereafter "CPK") to not fail when attempting to use a more recent version of the Gateway CRDs in Kind. If you spin up a fresh kind cluster today, then apply the v1.5.0 CRDs to that cluster, then attempt to run `cloud-provider-kind`, then CPK will fail due to a version mismatch. The only fix is to locally build a version of CPK which includes version 1.5.0 of the Gateway CRDs. Steps to reproduce using commit `3392ee06aac6fcac4817f92a1aafcdb9787a51f6` of CPK:

```
./kind create cluster
kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.5.0/standard-install.yaml
cloud-provider-kind
```

Fails with:

```
E0512 15:54:03.456491 1063399 controller.go:294] "Failed to install Gateway API CRDs" err="error processing embedded CRDs from crds/standard: failed to create CRD \"backendtlspolicies.gateway.networking.k8s.io\" from file \"crds/standard/gateway.networking.k8s.io_backendtlspolicies.yaml\": customresourcedefinitions.apiextensions.k8s.io \"backendtlspolicies.gateway.networking.k8s.io\" is forbidden: ValidatingAdmissionPolicy 'safe-upgrades.gateway.networking.k8s.io' with binding 'safe-upgrades.gateway.networking.k8s.io' denied request: In>
```

Additionally, a prior change has already upgraded the Go dependency for gateway-api to v1.5.1, implying that this upgrade should have already happened but may have been overlooked. Change was in #386

--------------

If this is not the right time to do this upgrade, or if this change is not in line with the contributing guidelines, then please close this PR. I'm just trying to be proactive about the version upgrades.

### DIFF
--- a/.github/workflows/gateway.yml
+++ b/.github/workflows/gateway.yml
@@ -7,7 +7,7 @@ on:
 
 env:
   GO_VERSION: "1.26"
-  GATEWAY_VERSION: "v1.4.1"
+  GATEWAY_VERSION: "v1.5.0"
   K8S_VERSION: "v1.35.0"
   KIND_VERSION: "v0.31.0"
   KIND_CLUSTER_NAME: "kind-cloud"
--- a/pkg/gateway/crds/experimental/gateway.networking.k8s.io_backendtlspolicies.yaml
+++ b/pkg/gateway/crds/experimental/gateway.networking.k8s.io_backendtlspolicies.yaml
@@ -2,8 +2,8 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    api-approved.kubernetes.io: https://github.com/kubernetes-sigs/gateway-api/pull/3328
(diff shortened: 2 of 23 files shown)


aojea:
awesome

/ok-to-test
/lgtm
/approve

Issue #387, opened 2026-03-17
`--enable-lb-port-mapping` exposes dual-stack host bindings, causing `localhost` connections to fail


gionn:
When running `cloud-provider-kind --enable-lb-port-mapping` (v0.10.0) on GitHub Actions runners, the `kindccm` container is started with host port bindings on both IPv4 and IPv6. In this environment, connecting to `localhost:<port>` can fail with connection resets, while forcing IPv4 with `curl -4` works.

This looks like an IPv6-related port forwarding problem in the GitHub Actions runner or Docker setup, where the host advertises an IPv6 listener even though the container or Docker daemon is not actually handling IPv6 traffic correctly.

### Findings

After starting `cloud-provider-kind --enable-lb-port-mapping`, the `kindccm` container is running:

```text
runner@runnervm46oaq:~/work$ docker ps -a
CONTAINER ID   IMAGE                      COMMAND                  CREATED          STATUS          PORTS                                                                                                                                     NAMES
70ee3a0370ba   envoyproxy/envoy:v1.33.2   "/docker-entrypoint.…"   9 minutes ago    Up 9 minutes    0.0.0.0:32768->80/tcp, [::]:32768->80/tcp, 0.0.0.0:32769->443/tcp, [::]:32769->443/tcp, 0.0.0.0:32770->10000/tcp, [::]:32770->10000/tcp   kindccm-1b469bea678e
1e2856c6f331   kindest/node:v1.34.3       "/usr/local/bin/entr…"   10 minutes ago   Up 10 minutes   127.0.0.1:38815->6443/tcp                                                                                                                 chart-testing-control-plane
```

The published service port is bound on both IPv4 and IPv6:

```text
runner@runnervm46oaq:~/work$ docker port 70ee3a0370ba 80/tcp
0.0.0.0:32768
[::]:32768
```

In pytest, connecting to `localhost:32768` fails with a connection reset:

```text
=================================== FAILURES ===================================
/opt/hostedtoolcache/Python/3.9.19/x64/lib/python3.9/site-packages/requests/adapters.py:501: requests.exceptions.ConnectionError: ('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))
=========================== short test summary info ============================
```

I can reproduce the same behavior with `curl`, unless I force IPv4 with `curl -4`.

Two workarounds seem to avoid the problem:

1. Force the client to use IPv4.
2. Disable IPv6 entirely on the runner:

```sh
[...]

PR #418, opened 2026-06-03
Allow a static host port number to be specified using `--enable-lb-port-mapping=<port>`

### LINKED ISSUE #417: Ability to specify static port when using `--enable-lb-port-mapping`
This was suggested a while back in https://github.com/kubernetes-sigs/cloud-provider-kind/issues/182 and while I agree that port conflicts could be undesirable, the dramatic increase in complexity required to dynamically retrieve the ephemeral port and configure it in dependent applications is not worth it for local development IMO.

I would really like to see a `--port` or `--publish` flag added for this. Would a PR be accepted?

As an aside, the `--enable-lb-port-mapping` flag did not seem to work for me at first and I found it very difficult to debug because it seems like the port information returned about the `kindccm` envoy container is incorrect:

```
docker ps

CONTAINER ID   IMAGE                      COMMAND                  CREATED          STATUS          PORTS                                                                    NAMES
4cebb83e06dc   envoyproxy/envoy:v1.33.2   "/docker-entrypoint.…"   41 minutes ago   Up 41 minutes   127.0.0.1:32784->80/tcp, 0.0.0.0:32785->10000/tcp, :::32785->10000/tcp   kindccm-gw-4eff45ecc9be
4671e3a7b274   kindest/node:v1.35.0       "/usr/local/bin/entr…"   42 minutes ago   Up 42 minutes   127.0.0.1:46517->6443/tcp                                                control-plane
```

Based on `127.0.0.1:32784->80/tcp`, I tried `curl localhost:32784` from the host but this does not work. 

However I eventually discovered that `docker port` reports different, seemingly more up to date information:

```
docker port kindccm-gw-4eff45ecc9be

80/tcp -> 0.0.0.0:65481
10000/tcp -> 0.0.0.0:32785
10000/tcp -> [::]:32785
```

And sure enough, `curl localhost:65481` does work....

jarpoole:
Closes https://github.com/kubernetes-sigs/cloud-provider-kind/issues/417

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -34,6 +34,9 @@ type Config struct {
 	// and do userspace proxying from the original port to the portmaps.
 	// If the cloud-provider-kind runs in a container on these platforms only enables portmapping.
 	LoadBalancerConnectivity Connectivity
+	// Host port used when LoadBalancerConnectivity is Portmap.
+	// 0 means the OS assigns an ephemeral port; any other value is used as-is.
+	LoadBalancerPortMappingHostPort int32
 	// Type of connectivity between the cloud-provider-kind and the clusters
 	ControlPlaneConnectivity Connectivity
 	// Gateway API Release channel (default stable)
--- a/pkg/gateway/envoy.go
+++ b/pkg/gateway/envoy.go
@@ -214,6 +214,7 @@ func createGateway(clusterName string, nameserver string, localAddress string, l
 		// advertised but not functional end-to-end (e.g. some GitHub Actions runners).
 		// For dual-stack or unspecified gateways, publish without an explicit address.
 		// See https://github.com/kubernetes-sigs/cloud-provider-kind/issues/387
+		hostPort := config.DefaultConfig.LoadBalancerPortMappingHostPort
 		seen := make(map[string]struct{})
 		for _, listener := range gateway.Spec.Listeners {
 			proto := "tcp"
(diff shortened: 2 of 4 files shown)


pkg/gateway/envoy.go
```diff
@@ -223,10 +224,10 @@ func createGateway(clusterName string, nameserver string, localAddress string, l
 
 			var publishArg string
 			if listenAddress != "" {
-				hostPortBinding := net.JoinHostPort(listenAddress, fmt.Sprintf("%d", listener.Port))
+				hostPortBinding := net.JoinHostPort(listenAddress, fmt.Sprintf("%d", hostPort))
```
jarpoole, line 227: @aojea I'm not 100% sure I understand what is going on here and I think there might be an issue related to https://github.com/kubernetes-sigs/cloud-provider-kind/issues/387. Even before this change, why is it that a port is explicitly specified but only if the host is not dual stack?

pkg/config/config.go
```diff
@@ -34,6 +34,9 @@ type Config struct {
 	// and do userspace proxying from the original port to the portmaps.
 	// If the cloud-provider-kind runs in a container on these platforms only enables portmapping.
 	LoadBalancerConnectivity Connectivity
+	// Host port used when LoadBalancerConnectivity is Portmap.
+	// 0 means the OS assigns an ephemeral port; any other value is used as-is.
+	LoadBalancerPortMappingHostPort int32
```
jarpoole, line 39:
```suggestion
	LoadBalancerPortMappingHostPort uint16
```

In other languages I would normally use `uint16` as this is precise for tcp ports but I didn't see any other usages of this type and I'm not super familiar with go so I figured it was best to follow the pattern

PR #422, opened 2026-06-08
feat: enable IPv6/dual-stack admin listener config for Envoy bootstrap

### LINKED ISSUE #421: [macOS + Colima] Envoy admin port is published dual-stack/random, but /ready is IPv4-only, causing undeterministic readiness failures on
### Description

On macOS with Colima, `cloud-provider-kind` randomly fails to bring up the `kindccm-*` load balancer container because the Envoy admin readiness check goes through a published host port that is exposed on both IPv4 and IPv6, while Envoy's admin listener is configured IPv4-only.

What I could verify:

- the Envoy admin listener inside the container is bound to `0.0.0.0:10000`, not `::`
  ```yaml
  root@kindccm-ac4c0bc3eb60:/# cat /home/envoy/envoy.yaml
  node:
    cluster: cloud-provider-kind
    id: cloud-provider-kind-id
  
  dynamic_resources:
    cds_config:
      resource_api_version: V3
      path: /home/envoy/cds.yaml
    lds_config:
      resource_api_version: V3
      path: /home/envoy/lds.yaml
  
  admin:
    access_log_path: /dev/stdout
    address:
      socket_address:
        address: 0.0.0.0
        port_value: 10000
  ```
- `curl http://127.0.0.1:10000/ready` works inside the container/VM
- `curl http://[::1]:10000/ready` does **not** work
- a standalone SSH forward to remote `0.0.0.0:<hostport>` works
- the same forward to remote `[::]:<hostport>` accepts the TCP connection but returns `Empty reply from server`

This seems to match the current implementation:

- `pkg/loadbalancer/proxy.go` configures Envoy admin on `0.0.0.0:10000`
- `pkg/loadbalancer/server.go` publishes the admin endpoint with `--publish=10000/tcp` and then `--publish-all`, so Docker exposes it on a random host port on both `0.0.0.0` and `::`
- `pkg/loadbalancer/proxy.go` then waits for readiness via `http://127.0.0.1:<mapped-admin-port>/ready`
- on macOS, `cmd/app.go` forces `LoadBalancerConnectivity = Tunnel`

So the likely culprit is that the admin port is published as a generic dual-stack/random port, but the readiness endpoint behind it is only actually usable over IPv4.

### Reproduction

1. `colima start --vm-type=vz --arch=aarch64 --vz-rosetta --mount-type=virtiofs --cpu 8 --memory=8 --disk=50`
2. `kind create cluster`
3. Run `sudo cloud-provider-kind` on macOS with Colima and create a `LoadBalancer` Service so that a `kindccm-*` container is created.
4. Observe the container port mappings: `docker ps && docker port kindccm-<id>`
5. Observe that ipv6 is exposed but `http://[::1]:<mapped-admin-port>/ready` does not work. In unlucky cirtumstances the SSH port mapping / routing is redirected to the Ipv6 port mapping which therefore cannot work.

tuunit:
Envoys admin port was always published as a generic dual-stack/random host port, while the admin listener itself was only safe on IPv4 unless IPv6 was actually enabled in the container. On macOS + Colima this could send /ready checks down an IPv6 path that accepted the TCP connection but returned an empty reply, which made load balancer startup flaky.

This change makes admin listener and port publishing consistent with actual IP-family support: IPv4-only setups stay IPv4-only, and IPv6-capable setups use a dual-stack Envoy admin listener plus dual-stack publishing.

Fixes #421

### DIFF
--- a/pkg/loadbalancer/proxy.go
+++ b/pkg/loadbalancer/proxy.go
@@ -48,8 +48,9 @@ admin:
   access_log_path: /dev/stdout
   address:
     socket_address:
-      address: 0.0.0.0
+      address: "::"
       port_value: 10000
+      ipv4_compat: true
 `
 
 // proxyConfigData is supplied to the loadbalancer config template
--- a/pkg/loadbalancer/server.go
+++ b/pkg/loadbalancer/server.go
@@ -246,9 +246,9 @@ func (s *Server) createLoadBalancer(clusterName string, service *v1.Service, ima
 			}
 		}
 	}
-	// publish the admin endpoint
+	// explicitely publish the admin endpoint
 	args = append(args, fmt.Sprintf("--publish=%d/%s", envoyAdminPort, v1.ProtocolTCP))
-	// Publish all ports in the host in random ports
+	// publish all ports in the host on random ports
 	args = append(args, "--publish-all")
 
 	if service.Spec.LoadBalancerIP != "" {


tuunit:
@stmcginnis can you check why the test report wasn't uploaded for the failed ipv4 test?

aojea:
```
2026-06-08T15:43:28.0920794Z Preparing 1 report as configured.
2026-06-08T15:43:28.4118156Z Kubernetes e2e suite:1 | [FAILED] expected error when trying to connect to cluster IP 39.489597067s
2026-06-08T15:43:28.4149202Z ℹ️ Posting with conclusion 'failure' to https://github.com/kubernetes-sigs/cloud-provider-kind/pull/422 (sha: 8737e60526c1ce6eb80fd2805c4352212898aa3b)
2026-06-08T15:43:28.4150906Z ##[endgroup]
2026-06-08T15:43:28.4152421Z ##[group]🚀 Publish results
2026-06-08T15:43:28.4154139Z ℹ️ - JUnit Test Report - 7184 tests run, 150 passed, 7033 skipped, 1 failed.
2026-06-08T15:43:28.4156425Z    🧪 - Kubernetes e2e suite | [FAILED] expected error when trying to connect to cluster IP
2026-06-08T15:43:28.4165377Z ℹ️ - JUnit Test Report - Creating check (Annotations: 1)
2026-06-08T15:43:28.5994795Z ##[error]❌ Failed to create checks using the provided token. (HttpError: Resource not accessible by integration - https://docs.github.com/rest/checks/runs#create-a-check-run)
2026-06-08T15:43:28.5999324Z ##[warning]⚠️ This usually indicates insufficient permissions. More details: https://github.com/mikepenz/action-junit-report/issues/23
2026-06-08T15:43:28.6017172Z ##[endgroup]
2026-06-08T15:43:28.6474816Z Post job cleanup.
```

image

pkg/gateway/envoy.go
```diff
@@ -256,3 +271,17 @@ func createGateway(clusterName string, nameserver string, localAddress string, l
...
+		return "0.0.0.0", false
+	}
+	return "::", true
+}
+
+func adminPortPublishArg(ipv6Enabled bool, port int) string {
+	if ipv6Enabled {
+		return fmt.Sprintf("--publish=%d/tcp", port)
+	}
+	return fmt.Sprintf("--publish=127.0.0.1::%d/tcp", port)
```
aojea:
```suggestion
	return fmt.Sprintf("--publish=127.0.0.1:%d/tcp", port)
```

?

pkg/gateway/envoy.go
```diff
@@ -78,14 +78,19 @@ admin:
   access_log_path: /dev/stdout
   address:
     socket_address:
-      address: 0.0.0.0
+      address: {{ if eq .AdminAddress "::" }}"::"{{ else }}{{ .AdminAddress }}{{ end }}
```
aojea: If we just always quote?
```suggestion
      address: "{{ .Address }}"
```

tuunit: I wanted to make it explicit and show that we are testing for ipv6 vs ipv4 but yes obviously the statement doesn't really make sense

pkg/gateway/envoy.go
tuunit: plain old typo sorry going to fix that interestingly docker just accepted it as is 🤔  now i'm curious how docker/mobby actually implemented the parsing

tuunit: i had a lot of problems with my local macos+colima setup so i tried a bunch of things. what we could do here as well is bind to `0.0.0.0:%d/tcp` rather than `127.0.0.1`. would be closer to the current behaviour. but i have to be honest i don't know if it would make a difference for `cloud-provider-kind`

tuunit:
> ```
> 2026-06-08T15:43:28.0920794Z Preparing 1 report as configured.
> 2026-06-08T15:43:28.4118156Z Kubernetes e2e suite:1 | [FAILED] expected error when trying to connect to cluster IP 39.489597067s


@aojea thanks for the quick feedback, highly appreciated!!

Just to make sure before I totally run into the wrong direction. The CI uses the official upstream conformance tests from the k8s repo, right?

https://github.com/kubernetes/kubernetes/tree/master/test/e2e

pkg/gateway/envoy.go
aojea: Then if we omit the IP address and we just leave the port as it was before, will not work in all cases?

tuunit: no because then docker publishes the port on both ipv6 and ipv4 (even if the docker daemon is only configured with ipv4). 

which we obviously do not want as this can cause a forwarding through something like colimas host agent to try to connect to the ipv6 route which will not yield a return.

See before:

```bash
$ docker ps
CONTAINER ID   IMAGE                      COMMAND                  CREATED         STATUS         PORTS                                                                                                                                     NAMES
212a55b089ff   envoyproxy/envoy:v1.33.2   "/docker-entrypoint.…"   3 seconds ago   Up 2 seconds   0.0.0.0:32768->80/tcp, [::]:32768->80/tcp, 0.0.0.0:32769->443/tcp, [::]:32769->443/tcp, 0.0.0.0:32770->10000/tcp, [::]:32770->10000/tcp   kindccm-ac4c0bc3eb60
b69dfe465710   kindest/node:v1.35.0       "/usr/local/bin/entr…"   24 hours ago    Up 7 minutes   127.0.0.1:54749->6443/tcp                                                                                                                 local-control-plane
```

After my change:

```bash
$ docker ps
CONTAINER ID   IMAGE                      COMMAND                  CREATED         STATUS         PORTS                                                                  NAMES
c2e8ce3c4a79   envoyproxy/envoy:v1.33.2   "/docker-entrypoint.…"   3 seconds ago   Up 2 seconds   0.0.0.0:80->80/tcp, 0.0.0.0:443->443/tcp, 127.0.0.1:32768->10000/tcp   kindccm-ac4c0bc3eb60
b69dfe465710   kindest/node:v1.35.0       "/usr/local/bin/entr…"   24 hours ago    Up 8 minutes   127.0.0.1:54749->6443/tcp                                              local-control-plane
```

tuunit: When creating a cluster using:

```bash
$ cat <<EOF > dual-stack.yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
networking:
  ipFamily: dual
  apiServerAddress: 127.0.0.1
nodes:
  - role: control-plane
EOF

$ kind create cluster --name dual --config=dual-stack.yaml

$ cat <<EOF > dual-svc.yaml
apiVersion: v1
kind: Service
metadata:
  name: dual-svc
spec:
  ipFamilyPolicy: RequireDualStack
  ports:
  - name: web
    port: 80
    protocol: TCP
    targetPort: web
  selector:
    app.kubernetes.io/name: my-app
  type: LoadBalancer
EOF

$ kubectl apply -f dual-svc.yaml

$ kubectl get svc
NAME         TYPE           CLUSTER-IP     EXTERNAL-IP                        PORT(S)        AGE
dual-svc     LoadBalancer   10.96.152.12   172.18.0.4,fc00:f853:ccd:e793::4   80:31519/TCP   5s
kubernetes   ClusterIP      10.96.0.1      <none>                             443/TCP        4m55s

$ docker ps
CONTAINER ID   IMAGE                      COMMAND                  CREATED          STATUS          PORTS                                                                                        NAMES
646db536e0f5   envoyproxy/envoy:v1.33.2   "/docker-entrypoint.…"   50 seconds ago   Up 50 seconds   0.0.0.0:32771->80/tcp, [::]:32771->80/tcp, 0.0.0.0:32772->10000/tcp, [::]:32772->10000/tcp   kindccm-0859a07e6515
0d2453a6903e   kindest/node:v1.36.1       "/usr/local/bin/entr…"   5 minutes ago    Up 5 minutes    127.0.0.1:53019->6443/tcp                                                                    dual-control-plane

Comment (tuunit) [reply pkg/gateway/envoy.go]: @aojea as you can see above, before this PR we alway expose ipv4 and ipv6 which can cause routing issues. With this change we expose ipv4 and ipv6 only if the cluster and service supports it and configure envoy accordingly

File: pkg/gateway/envoy.go [hunk 3]
```diff
@@ -237,6 +248,10 @@ func createGateway(clusterName string, nameserver string, localAddress string, l
 			args = append(args, publishArg)
 		}
 	}
+	// Publish the admin endpoint according to the gateway's IP-family support. IPv4-only
+	// gateways keep the admin port on IPv4 localhost, while gateways with IPv6 enabled
+	// publish the dual-stack-capable admin listener on both host families.
+	args = append(args, adminPortPublishArg(adminIPv4Compat, envoyAdminPort))
```
aojea: where was this published in gateway before?

aojea:
let's go back to the blackboard, I do not know if this is exactly the problem

tuunit:
> let's go back to the blackboard, I do not know if this is exactly the problem

The problem is actually easily stated:

Problem statement: 
The envoy admin endpoints (which are used for the readiness check) are only listening on IPv4 but the kindccm containers are exposing ipv4 and ipv6.

(Which is obviously a bug because curl'ing the readiness endpoint via ipv6 will fail)

```bash
❯ curl "http://127.0.0.1:32772/ready"
LIVE
❯ curl "http://[::1]:32772/ready"
curl: (7) Failed to connect to ::1 port 32772 after 0 ms: Couldn't connect to server
```

Solutions:
Option 1: Only expose ipv4 for the containers as well (small change)
Option 2: Configure envoy to listen on ipv4 & ipv6 as well based on the fact if dual-stack is active (which i opted for)


I might have overstepped my scope and changes, as I wanted to take the opportunity to introduce full dual-stack support for all endpoints. @aojea I would be up for a quick chat / huddle on Slack or Google meets as well

pkg/gateway/envoy.go
tuunit: It wasn't explicitly published with any settings but rather caught by `args = append(args, "--publish-all")` one line below. Therefore we didn't have any control over it.

aojea:
> The envoy admin endpoints (which are used for the readiness check)

What are "readiness checks" in this context?

tuunit:
> > The envoy admin endpoints (which are used for the readiness check)
> 
> What are "readiness checks" in this context?

cloud-provider-kind checks if the loadbalancer containers are up and only then attaches the internal ip on the loop device and starts listening on it

https://github.com/kubernetes-sigs/cloud-provider-kind/blob/78a4cf750a50fbf5fc69f2bb1b4599a1fd4c790a/pkg/loadbalancer/proxy.go#L341-L354

aojea:
yeah, but it always uses ipv4 https://github.com/kubernetes-sigs/cloud-provider-kind/blob/78a4cf750a50fbf5fc69f2bb1b4599a1fd4c790a/pkg/loadbalancer/proxy.go#L326-L339

and only is used for loadbalancer.

So I'm not arguing that we can not make it better but I was surprised you had seen underterministic readiness failures, because I do not see how can that be possible if the checks are always using ipv4

> causing undeterministic readiness failures on

tuunit:
> So I'm not arguing that we can not make it better but I was surprised you had seen underterministic readiness failures, because I do not see how can that be possible if the checks are always using ipv4
> 
> > causing undeterministic readiness failures on

It only becomes a problem on macos with colima because:


In case of colima when i run cloud-provider-kind on the macos host. It creates the docker container and then tries to check when the loadbalancer container `kindccm-*` is ready. Unfortunately as the container is exposed on ipv4 and ipv6. The routing from macos host goes through the ssh socket of colima and depending on which port-forwarding happens first (ipv4 vs. ipv6) it either works or it fails (non-deterministic).


cloud-provider-kind (on macos host) --> `http://127.0.0.1:<node-port>/ready` -> ssh socket file of colima ---- either ipv4 or ipv6 route ----> container


This part above is the non-deterministic issue and causes unexpected connection / readiness check failures. If cloud-provider-kind wouldn't start the container with ipv6 exposed or would listen correctly on both protocols this wouldn't be a problem.

In my opinion colima does the correct thing. It forwards all ports on both protocols that are defined / exposed / published by the docker deployment. 


Detailed log:

```
sudo cloud-provider-kind
Password:
I0610 15:20:20.787473   39830 app.go:77] FLAG: --enable-default-ingress="true"
I0610 15:20:20.787504   39830 app.go:77] FLAG: --enable-lb-port-mapping="false"
I0610 15:20:20.787507   39830 app.go:77] FLAG: --enable-log-dumping="false"
I0610 15:20:20.787509   39830 app.go:77] FLAG: --gateway-channel="standard"
I0610 15:20:20.787512   39830 app.go:77] FLAG: --help="false"
I0610 15:20:20.787514   39830 app.go:77] FLAG: --logs-dir=""
I0610 15:20:20.787516   39830 app.go:77] FLAG: --verbosity="2"
I0610 15:20:21.022492   39830 http.go:16] probe HTTP address https://127.0.0.1:53019
I0610 15:20:21.022503   39830 http.go:16] probe HTTP address https://dual-control-plane:6443
I0610 15:20:21.026129   39830 controller.go:169] "Connected succesfully" cluster="dual" host="https://127.0.0.1:53019"
I0610 15:20:21.026144   39830 controller.go:96] "Creating new cloud provider" cluster="dual"
...
I0610 15:20:21.031669   39830 controller.go:235] Starting service controller
I0610 15:20:21.031746   39830 shared_informer.go:349] "Waiting for caches to sync" controller="service"
...
I0610 15:20:21.269244   39830 server.go:96] container kindccm-0859a07e6515 for loadbalancer is not running
I0610 15:20:21.304858   39830 server.go:105] creating container for loadbalancer
I0610 15:20:21.304926   39830 server.go:257] creating loadbalancer with parameters: [--detach --tty --label io.x-k8s.cloud-provider-kind.cluster=dual --label io.x-k8s.cloud-provider-kind.loadbalancer.name=dual/defau
lt/dual-svc --net kind --init=false --hostname kindccm-0859a07e6515 --privileged --restart=on-failure --sysctl=net.ipv4.ip_forward=1 --sysctl=net.ipv4.conf.all.rp_filter=0 --sysctl=net.ipv4.ip_unprivileged_port_star
t=1 --sysctl=net.ipv6.conf.all.disable_ipv6=0 --sysctl=net.ipv6.conf.all.forwarding=1 --publish=80/TCP --publish=10000/TCP --publish-all docker.io/envoyproxy/envoy:v1.33.2 bash -c echo -en 'node:
  cluster: cloud-provider-kind
  id: cloud-provider-kind-id

dynamic_resources:
  cds_config:
    resource_api_version: V3
    path: /home/envoy/cds.yaml
  lds_config:
    resource_api_version: V3
    path: /home/envoy/lds.yaml

admin:
  access_log_path: /dev/stdout
  address:
    socket_address:
      address: 0.0.0.0
      port_value: 10000
' > /home/envoy/envoy.yaml && touch /home/envoy/cds.yaml && touch /home/envoy/lds.yaml && while true; do envoy -c /home/envoy/envoy.yaml && break; sleep 1; done]
I0610 15:20:21.324880   39830 tunnel.go:36] found port maps map[10000/tcp:32790 80/tcp:32789] associated to container kindccm-gw-8c4626fc4e39
I0610 15:20:21.336742   39830 tunnel.go:43] setting IPv4 address 172.18.0.3 associated to container kindccm-gw-8c4626fc4e39
I0610 15:20:21.336850   39830 tunnel.go:142] Starting tunnel on 172.18.0.3:80 tcp
...
I0610 15:20:21.609588   39830 proxy.go:342] unexpected error trying to get load balancer kindccm-0859a07e6515 readiness :Get "http://127.0.0.1:32792/ready": EOF
I0610 15:20:22.612669   39830 proxy.go:342] unexpected error trying to get load balancer kindccm-0859a07e6515 readiness :Get "http://127.0.0.1:32792/ready": EOF
I0610 15:20:23.611320   39830 proxy.go:342] unexpected error trying to get load balancer kindccm-0859a07e6515 readiness :Get "http://127.0.0.1:32792/ready": EOF
I0610 15:20:24.611704   39830 proxy.go:342] unexpected error trying to get load balancer kindccm-0859a07e6515 readiness :Get "http://127.0.0.1:32792/ready": EOF
I0610 15:20:25.611176   39830 proxy.go:342] unexpected error trying to get load balancer kindccm-0859a07e6515 readiness :Get "http://127.0.0.1:32792/ready": EOF
I0610 15:20:26.611827   39830 proxy.go:342] unexpected error trying to get load balancer kindccm-0859a07e6515 readiness :Get "http://127.0.0.1:32792/ready": EOF
I0610 15:20:27.612291   39830 proxy.go:342] unexpected error trying to get load balancer kindccm-0859a07e6515 readiness :Get "http://127.0.0.1:32792/ready": EOF
...
```

aojea:
> cloud-provider-kind (on macos host) --> `http://127.0.0.1:<node-port>/ready` -> ssh socket file of colima ---- either ipv4 or ipv6 route ----> container
> 
> This part above is the non-deterministic issue and causes unexpected connection / readiness check failures. If cloud-provider-kind wouldn't start the container with ipv6 exposed or would listen correctly on both protocols this wouldn't be a problem.
> 
> In my opinion colima does the correct thing. It forwards all ports on both protocols that are defined / exposed / published by the docker deployment.

please bear with me as there are many pieces here I'm trying to understand 

> http://127.0.0.1:<node-port>/ready` -> ssh socket file of colima ---- either ipv4 or ipv6 route ----> container

So colima is taking a packet in an ipv4 socket and randomly forwarding it via ipv4 or ipv6?

tuunit:
Thinking about it. The obvious choice might be to just not expose ipv6 as cloud-provider-kind always tries to call an ipv4 for the checks, right? 🤔 

Meaning we should probably just change the publish arg from `--publish=10000/tcp` to `--publish=0.0.0.0:10000/tcp`

tuunit:
> > cloud-provider-kind (on macos host) --> `http://127.0.0.1:<node-port>/ready` -> ssh socket file of colima ---- either ipv4 or ipv6 route ----> container
> > This part above is the non-deterministic issue and causes unexpected connection / readiness check failures. If cloud-provider-kind wouldn't start the container with ipv6 exposed or would listen correctly on both protocols this wouldn't be a problem.
> > In my opinion colima does the correct thing. It forwards all ports on both protocols that are defined / exposed / published by the docker deployment.
> 
> please bear with me as there are many pieces here I'm trying to understand
> 
> > http://127.0.0.1:/ready` -> ssh socket file of colima ---- either ipv4 or ipv6 route ----> container
> 
> So colima is taking a packet in an ipv4 socket and randomly forwarding it via ipv4 or ipv6?

No worries it took me like 6 hours to understand in full while debugging this mess...

From my understanding:

1. Lima has a guest agent running inside the VM watching for docker publish events.
   https://github.com/lima-vm/lima/tree/master/cmd/lima-guestagent
2. Colima maps the published ports on the MacOS host and points them to a SSH socket file:
   `.colima/_lima/colima/ssh.sock`
3. This way the some ports that get published inside the VM (Ubuntu based) via docker are available on the MacOS host and it feels like a natural integration (I think Docker Desktop and OrbStack must be doing something similar)

The tricky part is what happens if a port is accessible inside the VM via IPv4 and IPv6.
Colima tries to create a port-forwarding for both IPv4 and IPv6 on 0.0.0.0

```json
{"error":"failed to run [ssh -F /dev/null -o IdentityFile=\"/Users/<user>/.colima/_lima/_config/user\" -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o NoHostAuthenticationForLocalhost=yes -o PreferredAuthentications=publickey -o Compression=no -o BatchMode=yes -o IdentitiesOnly=yes -o GSSAPIAuthentication=no -o Ciphers=\"^aes128-gcm@openssh.com,aes256-gcm@openssh.com\" -o User=<user> -o ControlMaster=auto -o ControlPath=\"/Users/<user>/.colima/_lima/colima/ssh.sock\" -o ControlPersist=yes -T -O forward -L 0.0.0.0:32768:[::]:32768 -N -f -p 49291 127.0.0.1 --]: \"\": exit status 255","level":"warning","msg":"failed to set up forwarding tcp port 32768 (negligible if already forwarded)","time":"2026-06-09T08:57:19+02:00"}
```

aojea:
> The obvious choice might be to just not expose ipv6 as cloud-provider-kind always tries to call an ipv4 for the checks, right? 🤔

So, some history here, I wanted cloud provider kind to be super reliable, and dual stack has been always a pain to work with , more when you have to support multiple OSs , specially because there is no consistency as this issue is showing.
The more reliable in end users OSs is to use ipv4 because nobody has ipv6 only, but that does not mean current implementation is the ideal, since ipv6 only will not work.

Let's try to start over:
- can you point me to the colima code? something that receives an ipv4 packet and forward randomizing ip families is a bug IMHO and they are going to have more reports about this weird issues
- we need to revisit this healthcheck logic I've added time ago, but at first thought I do not see why we just not listen in both ipv6 and ipv4 always in the admin port (I don;t think we need to change the way we publish ports honestly)

tuunit:
> * we need to revisit this healthcheck logic I've added time ago, but at first thought I do not see why we just not listen in both ipv6 and ipv4 always in the admin port

what happens if ipv6 is deactivate for the docker daemon? will envoy be able to start listening on both? i would need to look into this!


> can you point me to the colima code? something that receives an ipv4 packet and forward randomizing ip families is a bug IMHO and they are going to have more reports about this weird issues

yes after writing it out in more detail this most definitely sounds like another bug on the co/lima side.

I did a little bit more research:

Colima is just the nice wrapper. Lima is the actual owner of the ssh socket:
https://lima-vm.io/docs/dev/internals/

https://github.com/lima-vm/lima/pull/1541

https://github.com/lima-vm/lima/blob/9b3195b4c236a8c02fbeeb8afca4123e25de5c9f/pkg/limatype/filenames/filenames.go#L55-L56

https://github.com/lima-vm/lima/blob/9b3195b4c236a8c02fbeeb8afca4123e25de5c9f/pkg/hostagent/hostagent.go#L249-L279

https://github.com/lima-vm/lima/blob/9b3195b4c236a8c02fbeeb8afca4123e25de5c9f/pkg/hostagent/port.go#L63-L96

I'll investigate this in more detail and will open an issue or even another PR with lima I guess

tuunit:
So how would you like to proceed with this PR? I still think it is wrong to publish the admin port on ipv6 if envoy only listens on ipv4. I'll be off for the day, thanks for the quick back and forth. I really appreciate it :)

aojea:
> if envoy only listens on ipv4.

let's listen always in all families on the admin port, that simplifies things a lot and is the default way of listening in golang https://github.com/golang/go/issues/17615 it seems c++ or envoy are more correct about the meaning of `0.0.0.0`

> I do not see why we just not listen in both ipv6 and ipv4 always in the admin port

tuunit:
@aojea what do you think about this way smaller scope of a change?

pkg/gateway/envoy.go
```diff
@@ -237,6 +239,9 @@ func createGateway(clusterName string, nameserver string, localAddress string, l
 			args = append(args, publishArg)
 		}
 	}
+	// explicitely publish the admin endpoint
+	args = append(args, fmt.Sprintf("--publish=%d/%s", envoyAdminPort, v1.ProtocolTCP))
```
aojea: can you please remind me why we need this? I prefer to not expose what is not necessary

pkg/loadbalancer/server.go
```diff
@@ -279,12 +281,8 @@ func isIPv6Service(service *v1.Service) bool {
...
 		return false
 	}
-	for _, family := range service.Spec.IPFamilies {
-		if family == v1.IPv6Protocol {
-			return true
-		}
-	}
-	return false
+
+	return slices.Contains(service.Spec.IPFamilies, v1.IPv6Protocol)
```
aojea: this is unrelated, let's not conflate changes

pkg/loadbalancer/proxy.go
```diff
@@ -48,8 +48,9 @@ admin:
   access_log_path: /dev/stdout
   address:
     socket_address:
-      address: 0.0.0.0
-      port_value: 10000
+      address: "::"
+      port_value: %d
+      ipv4_compat: true
```
aojea, line 53: this is unrelated too, let's not conflate changes

pkg/loadbalancer/server.go
tuunit: I follow the boyscout rule while programming. When I see a small code smell or improvement I usually just apply it.

pkg/loadbalancer/proxy.go
tuunit: what exactly is unrelated here? Isn't that exactly what we want to fix? supporting both ipv6 and ipv4

pkg/loadbalancer/server.go
tuunit: but of course i can revert the change

pkg/gateway/envoy.go
tuunit: the line below `--publish-all` already published every available port defined as part of the container which includes the admin port anyways

but it is cleaner to be explicit like done in `pkg/proxy/proxy.go`

and we need it because this is the port used by cloud-provider-kind if the container is ready and reachable

pkg/loadbalancer/server.go
aojea: but in shared code bases it is better to use independent commits, for any reason those "improvements" can cause a bug and change in behavior and then is hard to bisected

pkg/loadbalancer/proxy.go
aojea: `%d` is new, same as https://github.com/kubernetes-sigs/cloud-provider-kind/pull/422#discussion_r3442674517, the diff is larger because of the %d addition that requires to now process the constant , if we do the  `supporting both ipv6 and ipv4` only this should be only a few lines change

pkg/gateway/envoy.go
aojea: > and we need it because this is the port used by cloud-provider-kind if the container is ready and reachable

that is only for the loadbalancer, the gateway envoy operates differently , let's keep the change focused

tuunit:
@aojea sorry for the wait. I now removed the gateway related changes and the admin port templating. This way its really just the minimal change for adding ipv4_compat and exposing on "::" as well as a minor comment change

aojea:
/lgtm
/approve

Thanks

PR #431, opened 2026-06-25
Error if no supported runtime is found

### LINKED ISSUE #430: v0.11 doesn't fail if there is no docker
Hello,

It seems that `DetectRuntime`  return `docker` by default even if there is no `docker`, command doesn't failed and stuck infinite loop.
Is this an intended changes? and can you take a look?

Thanks,

stmcginnis:
We made the (not unreasonable) assumption that there would be a container runtime available when executing, and default to reporting docker if no other runtime could explicitly be detected.

If we did execute on a system without an available runtime, this would cause the unintended side effect that we would retry forever, treating the lack of a runtime as a transient error that should resolve eventually. This may or may not be what the user expects.

This changes the behavior to just error out of there is a runtime detection problem. It's likely the user will need to resolve whatever issue they have with their environment anyway, after which they can just rerun the command.

Closes: #430

### DIFF
--- a/cmd/app.go
+++ b/cmd/app.go
@@ -199,7 +199,11 @@ func runE(cmd *cobra.Command, args []string) error {
 		}
 	}
 	if option == nil {
-		p := container.DetectRuntime()
+		p, err := container.DetectRuntime()
+		if err != nil {
+			return err
+		}
+
 		container.SetRuntime(p)
 		switch p {
 		case "podman":
--- a/pkg/container/container.go
+++ b/pkg/container/container.go
@@ -49,22 +49,23 @@ func SetRuntime(name string) {
 }
 
 // DetectRuntime probes the system for an available container runtime.
-func DetectRuntime() string {
+// It returns an error if no supported runtime is installed and running.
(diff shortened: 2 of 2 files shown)


aojea:
I think that probably we implemented kind logic (cc @BenTheElder ), and on the contrary to kind, cloud-provider-kind keeps running checking for the clusters, is that correct?

stmcginnis:
In kind we will [always return a provider](https://github.com/kubernetes-sigs/kind/blob/42ef995e7631397b46a4dc855e93fb5443c25eef/pkg/cluster/internal/providers/docker/provider.go#L42), whether it is actually available or not. But then in most cases we will error when it comes time to [running a command](https://github.com/kubernetes-sigs/kind/blob/42ef995e7631397b46a4dc855e93fb5443c25eef/pkg/cluster/internal/providers/docker/provider.go#L108). I think the big difference there is that we don't have retries that treat these errors as transient.

aojea:
/lgtm
/approve


so we need a new 0.11.1 then

PR #438, opened 2026-07-06
Bump  Gateway CRDs to v1.6.0

### LINKED ISSUE #437: Update Gateway CRDs to 1.6.0
We need to update the CRDs too https://github.com/kubernetes-sigs/cloud-provider-kind/tree/main/pkg/gateway/crds

_Originally posted by @aojea in https://github.com/kubernetes-sigs/cloud-provider-kind/issues/435#issuecomment-4892086114_

/help

anup-kodlekere:
Closes #437

### DIFF
--- a/.github/workflows/gateway.yml
+++ b/.github/workflows/gateway.yml
@@ -7,7 +7,7 @@ on:
 
 env:
   GO_VERSION: "1.26"
-  GATEWAY_VERSION: "v1.5.0"
+  GATEWAY_VERSION: "v1.6.0"
   K8S_VERSION: "v1.35.0"
   KIND_VERSION: "v0.31.0"
   KIND_CLUSTER_NAME: "kind-cloud"
@@ -110,7 +110,7 @@ jobs:
 
     - name: Run tests
       run: |
-        /usr/local/bin/conformance.test -test.run TestConformance \
+        /usr/local/bin/conformance.test -test.run 'TestConformance$' \
             --debug=true \
             --kubeconfig=_artifacts/kubeconfig.conf \
             --organization=sigs.k8s.io \
--- a/pkg/gateway/crds/experimental/gateway.networking.k8s.io_backendtlspolicies.yaml
+++ b/pkg/gateway/crds/experimental/gateway.networking.k8s.io_backendtlspolicies.yaml
@@ -3,7 +3,7 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     api-approved.kubernetes.io: https://github.com/kubernetes-sigs/gateway-api/pull/4530
-    gateway.networking.k8s.io/bundle-version: v1.5.0
+    gateway.networking.k8s.io/bundle-version: v1.6.0
     gateway.networking.k8s.io/channel: experimental
   labels:
     gateway.networking.k8s.io/policy: Direct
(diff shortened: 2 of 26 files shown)


anup-kodlekere:
Okay, so apart from the CLA (I'll get that fixed) it looks like the gateway tests are failing. 
https://github.com/kubernetes-sigs/cloud-provider-kind/actions/runs/28799622885/job/85399131930?pr=438#step:9:29
I checked the runs from the last CRD update and it seems this particular step didn't run? 
https://github.com/kubernetes-sigs/cloud-provider-kind/actions/runs/25769202325/job/75688446421#step:9:25

.github/workflows/gateway.yml
stmcginnis: v1.6.0 added a set of `TestConformanceOptions_*` that this is now prefix matching on, so I think we just need to update this to only run the `TestConformance` tests we were expecting:

```suggestion
        /usr/local/bin/conformance.test -test.run 'TestConformance$' \
```

stmcginnis (commented):
Looks like we will need some test updates to support v1.6.0 fully. Thanks for a little LLM help, here's the list of updates I see we need.

**GatewayListenerUnsupportedProtocol**

We currently set every listener to default to `Accepted=True`, but this new test expects unsupported protocols to return `Accepted=False/ListenersNotValid`. This is in `gateway.go`.

**GatewayInvalidParametersRef**

Similar to above, this test now expects to get `Accetped=False/InvalidParameters` with an invalid `parametersRef`. We don't currently have any validation for this on our side.

**HTTPRouteNoBackedRefs**

Validates rules with omitted or emtpy `backendRefs` respond with a 500 and route has `ResolvedRefs=True`. Needs updates in `httproute.go`.

anup-kodlekere:
Thanks for the additional insights! I'm not sure I fully follow - but I would still like to take this ahead and attempt a fix. :)

stmcginnis:
> Thanks for the additional insights! I'm not sure I fully follow - but I would still like to take this ahead and attempt a fix. :)

With the bump to v1.6.0 in the GitHub Action workflow, that pulls in the newer conformance tests from https://github.com/kubernetes-sigs/gateway-api/tree/main/conformance. Those conformance tests have more test coverage and other updates, so even though we now have the newer v1.6.0 CRDs with this change, there are behaviors in the code that are being tested that don't quite match with what the conformance test expects.

So we have two options. We can update our behavior (hopefully just in the two files I referenced above) and get everything fully up to Gateway API v1.6.0. Or we can break this into two steps. This PR could just update the CRDs and we can leave the GitHub Action workflow pinned to v1.5.0. That would mean the CI testing would still use the older conformance binary and we _should_ be passing on that since that's what we've been running against so far. Then we would need to do a follow on PR to fully move us up to pass the expected conformance behavior for v.1.6.0.

I think we should just address the issues now and get everything updated. It does mean a little more work, but hopefully that's a good way to get a little more into the code. But that's definitely a bigger effort than just regenerating CRDs.

anup-kodlekere:
Thank you Sean! If this isn't too critical, I'm up for going through the code and attempt the changes in the files you mentioned before. If this is something that needs an immediate fix, then I can change the workflow to use the older Gateway version back to 1.5.0

aojea:
> Thank you Sean! If this isn't too critical, I'm up for going through the code and attempt the changes in the files you mentioned before. If this is something that needs an immediate fix, then I can change the workflow to use the older Gateway version back to 1.5.0

let's move everything to 1.6.0 in lockstep, there is no immediate urgency 

/ok-to-test
@anup-kodlekere remember to sign the CLA

aojea:
/close

thanks for the work but this was taking over by other contributor 

https://github.com/kubernetes-sigs/cloud-provider-kind/pull/447

PR #440, opened 2026-07-16
Reject HTTPRoute rules with unknown filters

### LINKED ISSUE #439: HTTPRoute with unsupported filters not rejected
According to https://gateway-api.sigs.k8s.io/reference/api-spec/main/spec/#httprouterule an implementation should set `Accepted` condition to `false` if an unsupported filter is encountered. However, the following HTTPRoute is accepted on v0.11.1:

```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  name: hello-rest
spec:
  parentRefs:
    - name: hello-rest
  rules:
    - filters:
        - type: ExternalAuth
          externalAuth:
            protocol: GRPC
            backendRef:
              kind: Service
              name: my-ext-authz
              port: 8080
            grpc:
              allowedHeaders:
                - Authorization
      backendRefs:
        - name: hello-rest
          port: 8080
```

status is as follows:

```yaml
status:
  parents:
  - conditions:
    - lastTransitionTime: "2026-07-13T04:51:00Z"
      message: Route is accepted.
      observedGeneration: 1
      reason: Accepted
      status: "True"
      type: Accepted
    - lastTransitionTime: "2026-07-13T04:51:00Z"
      message: All references resolved
      observedGeneration: 1
      reason: ResolvedRefs
      status: "True"
      type: ResolvedRefs
    controllerName: kind.sigs.k8s.io/gateway-controller
    parentRef:
      group: gateway.networking.k8s.io
      kind: Gateway
      name: hello-rest
```

Since `ExternalAuth` is not implemented, should cloud-provider-kind not reject this HTTPRoute? The case of external-auth seems particularly problematic, since the current behavior will let through an unauthenticated request to a service that was expecting to be protected.

bittrance:
cloud-provider-kind previously skipped filters with unknown types, leading to fail-open scenarios (particularly problematic with ExternalAuth). Such rules are now rejected. This also means HTTPRoutes can get condition PartiallyInvalid.

This PR only updates HTTPRoute as it seems GRPCRoute does not handle filters at all?

References:
- https://gateway-api.sigs.k8s.io/reference/api-spec/main/spec/#httproutefilter
- https://github.com/kubernetes-sigs/gateway-api/blob/5b010680508ea68f62bfe1a6302245eb796be7b5/apis/v1/shared_types.go#L444

Closes #439 .


aojea:
@bittrance it seems we need to update the gateway API and its associated tests to 1.6 to validate these changes https://github.com/kubernetes-sigs/cloud-provider-kind/pull/438 , it seems that PR got stale and the contributor didn't sign the CLA so we can not add him as coauthor.

Are you able to add a new commit to bump gateway api code and conformance test to 1.6 so we can validate these changes?

pkg/gateway/httproute.go
```diff
@@ -147,7 +155,35 @@ func translateHTTPRouteToEnvoyRoutes(
...
 	}
-	return envoyRoutes, allValidBackendRefs, overallCondition
+
+	invalidRuleCount := len(droppedRuleMessages)
+	if invalidRuleCount > 0 && invalidRuleCount == totalRules {
+		msg := fmt.Sprintf("no rules could be translated: %s", strings.Join(droppedRuleMessages, "; "))
+		return nil, nil, []metav1.Condition{
+			createFailureCondition(gatewayv1.RouteReasonUnsupportedValue, msg, httpRoute.Generation),
+		}
+	}
```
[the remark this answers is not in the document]
snorwin: I agree with Copilot here. I'm also wondering whether this applies elsewhere, e.g., in `translateHTTPRouteMatch`, I think we should be setting the `Accepted` condition to UnsupportedValue rather than `ResolvedRefs`.

pkg/gateway/httproute.go
```diff
@@ -18,18 +18,26 @@ import (
 )
 
 // translateHTTPRouteToEnvoyRoutes translates a full HTTPRoute into a slice of Envoy Routes.
-// It now correctly handles RequestHeaderModifier filters.
 func translateHTTPRouteToEnvoyRoutes(
 	httpRoute *gatewayv1.HTTPRoute,
 	serviceLister corev1listers.ServiceLister,
 	referenceGrantLister gatewaylistersv1.ReferenceGrantLister,
-) ([]*routev3.Route, []gatewayv1.BackendRef, metav1.Condition) {
+) ([]*routev3.Route, []gatewayv1.BackendRef, []metav1.Condition) {
```
snorwin: IMO returning arbitrary conditions as `[]metav1.Condition` is quite confusing and makes it hard to follow what's actually happening. I would prefer returning named `*metav1.Condition` values for the specific conditions instead.

pkg/gateway/httproute_test.go
```diff
@@ -0,0 +1,273 @@
...
+			wantResolvedRefsStatus: metav1.ConditionTrue,
+			wantPartiallyInvalid:   true,
+		},
+		{
+			name: "mix of supported and unsupported across three rules",
+			rules: []gatewayv1.HTTPRouteRule{
+				makeRuleWithFilters(redirectFilter()),
+				makeRuleWithFilters(urlRewriteFilter()),
+				makeRuleWithFilters(responsHeaderFilter()),
+			},
```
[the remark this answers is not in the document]
bittrance: Fixed.

pkg/gateway/httproute_test.go
```diff
@@ -0,0 +1,273 @@
...
+			wantResolvedRefsStatus: metav1.ConditionFalse,
+			wantPartiallyInvalid:   false,
+		},
+		{
+			name: "all rules have unsupported filters - fully invalid",
+			rules: []gatewayv1.HTTPRouteRule{
+				makeRuleWithFilters(urlRewriteFilter()),
+				makeRuleWithFilters(responsHeaderFilter()),
+			},
+			wantRoutes:             0,
```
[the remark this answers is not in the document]
bittrance: Not sure why copilot would want to remove `wantRoutes` here. It is admittedly default value, but it does not complain about the case above and removing it hardly improves clarity.

Rename fixed.

pkg/gateway/httproute_test.go
```diff
@@ -0,0 +1,273 @@
...
+		URLRewrite: &gatewayv1.HTTPURLRewriteFilter{},
+	}
+}
+
+func responsHeaderFilter() gatewayv1.HTTPRouteFilter {
+	return gatewayv1.HTTPRouteFilter{
+		Type:                   gatewayv1.HTTPRouteFilterResponseHeaderModifier,
+		ResponseHeaderModifier: &gatewayv1.HTTPHeaderFilter{},
+	}
+}
```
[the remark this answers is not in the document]
bittrance: Fixed.

pkg/gateway/httproute.go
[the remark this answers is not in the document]
bittrance: At some point, I started thinking the `createFailureCondition` helper was creating Accepted conditions, but actually, I now think this goes deeper. I'm looking at https://github.com/kubernetes-sigs/cloud-provider-kind/blob/1f390bc926365442853211eb42e14ba1da534850/pkg/gateway/httproute.go#L239 which also seems to assume the same thing. According to https://github.com/kubernetes-sigs/gateway-api/blob/74951f4ee876e53af857d3f8e2ba940a5cee5469/apis/v1/shared_types.go, ResolvedRef cannot have UnsupportedValue. A more thorough analysis of the conditions generation might be in order?

pkg/gateway/httproute.go
youngnick: Agreed. There will be a limited number of standard conditions (`Accepted`, `ResolvedRoutes`, and `Programmed` off the top of my head), and it's unlikely that we'll add more without a _very_ good reason.

pkg/gateway/httproute.go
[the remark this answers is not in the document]
youngnick: Yeah, `ResolvedRefs` is supposed to reflect the state of _references_ to other objects. So if all the objects referenced in the Route exist and are readable (not missing a ReferenceGrant or something), then `ResolvedRefs` should be `true`. That's why it doesn't support `UnsupportedValue`.

`Accepted` is the right place to put `UnsupportedValue`, but I should note that `Accepted` should only be `false` when there are _zero_ valid Rules in the Route. If there is a Route with an unsupported Filter in one Rule, and a valid Rule in another, then the Route should be `Accepted`, with _partial_ validity (and the valid Rule programmed, while the invalid, `UnsupportedValue` one is not).

We almost certainly don't test this in conformance properly, and I'm not sure which Reason we should use in that case.

pkg/gateway/httproute.go
bittrance: Fixed. Now returning three specific `*metav1.Condition` (Accepted, ResolvedRefs and PartiallyInvalid), but only if they indicate failures.

pkg/gateway/httproute_test.go
```diff
@@ -0,0 +1,273 @@
...
+			condByType := make(map[string]metav1.Condition)
+			for _, c := range conditions {
+				condByType[c.Type] = c
+			}
+
+			// Check ResolvedRefs condition.
+			resolvedRefs, ok := condByType[string(gatewayv1.RouteConditionResolvedRefs)]
+			if !ok {
+				t.Fatalf("ResolvedRefs condition missing from returned conditions")
+			}
```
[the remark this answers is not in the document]
bittrance: Fixed.

pkg/gateway/httproute.go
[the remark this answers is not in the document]
bittrance: Fixed, I think.

bittrance:
I **think** I have now met reviewer comments so far. The main change is to have `translateHTTPRouteToEnvoyRoutes` and its helpers focus exclusively on adverse conditions. The condition helpers are renamed to describe more precisely what they do. The helpers was also used by GRPCRoute; I took the liberty of deleting it rather than spend effort maintaining it, since it would need fixes similar to the ones in this PR. Tell me if this is not ok.

@aojea I don't think these changes are affected by 1.5 -> 1.6 migration. This branch passes the 1.5 conformance tests, but as youngnick points out, the tests are not watertight when it comes to reported conditions.

I don't see how I am responsible for the IPv6 failure, but perhaps someone knows what is going on?

pkg/gateway/gateway.go
```diff
@@ -250,14 +250,23 @@ func (c *Controller) buildEnvoyResourcesForGateway(gateway *gatewayv1.Gateway) (
...
 					key := types.NamespacedName{Name: httpRoute.Name, Namespace: httpRoute.Namespace}
 					currentParentStatuses := httpRouteStatuses[key]
 					for i := range currentParentStatuses {
-						// Only add the ResolvedRefs condition if the parent was Accepted.
+						if notAccepted != nil {
+							meta.SetStatusCondition(&currentParentStatuses[i].Conditions, *notAccepted)
+						}
 						if meta.IsStatusConditionTrue(currentParentStatuses[i].Conditions, string(gatewayv1.RouteConditionAccepted)) {
-							meta.SetStatusCondition(&currentParentStatuses[i].Conditions, resolvedRefsCondition)
+							if resolvedRefsFailure != nil {
```
snorwin, line 262: nit: could we simplify this by reverting to having `translateHTTPRouteToEnvoyRoutes` always return the `ResolvedRefs` condition

pkg/gateway/gateway.go
```diff
@@ -250,14 +250,23 @@ func (c *Controller) buildEnvoyResourcesForGateway(gateway *gatewayv1.Gateway) (
...
+								meta.SetStatusCondition(&currentParentStatuses[i].Conditions, *resolvedRefsFailure)
+							} else {
+								meta.SetStatusCondition(&currentParentStatuses[i].Conditions, createResolvedCondition(httpRoute.Generation))
+							}
+							if partiallyInvalidCond != nil {
+								meta.SetStatusCondition(&currentParentStatuses[i].Conditions, *partiallyInvalidCond)
```
snorwin, line 268: Will the `PartiallyInvalid` condition be removed again once the issue is fixed?

pkg/gateway/httproute.go
```diff
@@ -18,18 +18,30 @@ import (
...
+//   - notAccepted: non-nil (Accepted=False) when the entire route is rejected
+//   - resolvedRefsFailure: non-nil (ResolvedRefs=False) only when a backend ref could not be resolved;
+//     nil signals success and the caller is responsible for setting ResolvedRefs=True
+//   - partiallyInvalid: non-nil (PartiallyInvalid=True) when some rules were dropped
 func translateHTTPRouteToEnvoyRoutes(
 	httpRoute *gatewayv1.HTTPRoute,
 	serviceLister corev1listers.ServiceLister,
 	referenceGrantLister gatewaylistersv1.ReferenceGrantLister,
-) ([]*routev3.Route, []gatewayv1.BackendRef, metav1.Condition) {
+) ([]*routev3.Route, []gatewayv1.BackendRef, *metav1.Condition, *metav1.Condition, *metav1.Condition) {
```
snorwin: nit: consider using named return parameters here, I think it would improve readability

snorwin (commented):
@bittrance thanks for incorporating the feedback.

Overall this looks good, I think the reasons and conditions are now set according to the Gateway API spec. However, I'm not sure if the conditions are properly reset, e.g., is the `PartiallyInvalid` condition removed when a route switches to `Accepted=false`, or when the issue causing the invalidity is resolved?

I've added some nits as well, but I usually don't work on this project, feel free to disregard any that don't make sense.

pkg/gateway/gateway.go
bittrance: My reasoning was that this way all the condition return values have the same contract. If `translateHTTPRouteToEnvoyRoutes` always returned conditions, the other two checks would have to inspect the returned condition to know what to do.

pkg/gateway/gateway.go
bittrance: According to a quick manual test, this seems to be working correctly. Presumably because we do this: https://github.com/kubernetes-sigs/cloud-provider-kind/blob/493435b3a233011bace2361cd4e4e3073708f8a5/pkg/gateway/gateway.go#L455 .

pkg/gateway/httproute.go
bittrance: Indeed. Fixed.

pkg/gateway/gateway.go
snorwin: @bittrance thanks for taking the time to verify this, really appreciated.

aojea:
/lgtm
/approve

Thanks

PR #447, opened 2026-07-28
Conform to Gateway API v1.6.1

### LINKED ISSUE #437: Update Gateway CRDs to 1.6.0

bittrance:
Since #438 seems to have stalled, here is a competing PR, including updates so that we pass relevant conformance tests.

Closes: #437


.github/workflows/gateway.yml
```diff
@@ -7,7 +7,7 @@ on:
 
 env:
   GO_VERSION: "1.26"
-  GATEWAY_VERSION: "v1.5.0"
+  GATEWAY_VERSION: "v1.6.0"
```
snorwin:
```suggestion
  GATEWAY_VERSION: "v1.6.1"
```

.github/workflows/gateway.yml
snorwin, line 123: Note: since GEP-2162 (https://gateway-api.sigs.k8s.io/geps/gep-2162/) graduated in v1.4.0, setting supported features manually is discouraged, they're inferred automatically from the GatewayClass status, provided they're set there correctly.

pkg/gateway/gateway.go
```diff
@@ -45,6 +45,19 @@ var (
 	)
 )
 
+// isOurGateway returns true when the Gateway's GatewayClass is managed by this
+// controller (i.e. its spec.controllerName matches ours). Using the controller
+// name rather than the hard-coded GWClassName constant means the controller
+// also handles Gateways that reference conformance-test-created classes (which
+// carry our controller name but a different class name).
```
snorwin, line 52: nice catch!

snorwin (commented):
@bittrance really nice work, thanks!

I've added some comments. I'm aware the one about supported features is unrelated to this change, it might be worth covering in a follow-up PR.

pkg/gateway/gateway.go
```diff
@@ -65,11 +78,36 @@ func (c *Controller) syncGateway(ctx context.Context, key string) error {
...
+	if gw.Spec.Infrastructure != nil && gw.Spec.Infrastructure.ParametersRef != nil {
+		newGw := gw.DeepCopy()
+		meta.SetStatusCondition(&newGw.Status.Conditions, metav1.Condition{
+			Type:               string(gatewayv1.GatewayConditionAccepted),
+			Status:             metav1.ConditionFalse,
+			Reason:             string(gatewayv1.GatewayReasonInvalidParameters),
+			Message:            "infrastructure.parametersRef is not supported",
+			ObservedGeneration: newGw.Generation,
+		})
+		meta.SetStatusCondition(&newGw.Status.Conditions, metav1.Condition{
```
snorwin: Maybe this can be handled more generally, since IMO a Gateway can never have `Programmed=true` if `Accepted=false`.

pkg/gateway/gateway.go
bittrance: There was a Slack-discussion about this being a common issue happening exactly when I was looking at the code :).

pkg/gateway/gateway.go
bittrance: Fixed. I think this is an improvement, but it made a rather large change.

.github/workflows/gateway.yml
bittrance: Our GatewayClass currently does not present supported features, so this sounds like a separate PR?

.github/workflows/gateway.yml
bittrance: Fixed.

bittrance:
@snorwin Comments met. The Gateway Accepted/Programmed rework got a bit large, but now the intention reads clearer, I think.

aojea:
/approve

/assign @snorwin 

for final lgtm

wow, this is awesome, you just didn;t fix only a bug you improved the codebase by 100

Thanks

.github/workflows/gateway.yml
snorwin: yes, it would be great if you could open a follow-up PR.

bittrance: It is in progress. Should be up tonight unless there are complications.

pkg/gateway/gateway.go
```diff
@@ -65,61 +78,82 @@ func (c *Controller) syncGateway(ctx context.Context, key string) error {
...
-		return fmt.Errorf("failed to ensure gateway container %s: %w", containerName, err)
-	}
+	if gw.Spec.Infrastructure != nil && gw.Spec.Infrastructure.ParametersRef != nil {
+		meta.SetStatusCondition(&newGw.Status.Conditions, metav1.Condition{
+			Type:               string(gatewayv1.GatewayConditionAccepted),
+			Status:             metav1.ConditionFalse,
+			Reason:             string(gatewayv1.GatewayReasonInvalidParameters),
+			Message:            "infrastructure.parametersRef is not supported",
+			ObservedGeneration: newGw.Generation,
+		})
```
snorwin, line 100: Would it make more sense for this to be part of `setAcceptedCondition`?

pkg/gateway/gateway.go
```diff
@@ -65,61 +78,82 @@ func (c *Controller) syncGateway(ctx context.Context, key string) error {
...
+	} else {
```
snorwin, line 101: In addition, the entire `else { ... }` branch (i.e., creating the container, assigning addresses, ...) should IMO only be executed once the Gateway becomes `Accepted=true`.

I would suggest structuring it as follows:
```go
		envoyResources, listenerStatuses, httpRouteStatuses, grpcRouteStatuses = c.buildEnvoyResourcesForGateway(newGw)
		newGw.Status.Listeners = listenerStatuses

		setAcceptedCondition(newGw)

		if meta.IsStatusConditionTrue(newGw.Status.Conditions, string(gatewayv1.GatewayConditionAccepted)) {
				// create the container, ...

				// update xds
		}

		setProgrammedCondition(newGw, xdsErr)
```

I'm not sure whether this breaks anything, since `newGw` gets mutated a lot along the way, but I think it would be worth a try.

snorwin (commented):
@bittrance thanks for incorporating the feedback.

I just had one further idea on how we could improve this, let me know what you think.

pkg/gateway/gateway.go
bittrance: I don't think so. I think the parameterRef check can either be seen as validation of the Gateway object, or as the logic to process `infrastructure`. If anything, it should be its own function, but then we would need to pick one of those alternatives. I'm thinking we dodge that question until we either 1) get more validation or 2) need infrastructure config params?

pkg/gateway/gateway.go
bittrance: This is a significant behavioral change. I think this would mean we switch from "your config sucks so I'm going to deprogram Envoy" to "your config sucks so I'm going to keep whatever was there before" (at least as far as `syncGateways` is concerned)? In my high-volume production setup, I would appreciate last-known-good, but in my local dev env, I might appreciate clear feedback that my config sucks.

I think this merits its own PR. I have a branch on the change here https://github.com/bittrance/cloud-provider-kind/tree/retain-previous-config and it passes conformance tests, so it seems to work. However, I don't think GitHub supports stacked cross-repo PRs, so we probably need to merge this one first?

bittrance:
Supported Features PR is here https://github.com/kubernetes-sigs/cloud-provider-kind/pull/448 .

pkg/gateway/gateway.go
snorwin: I fully agree with you. However, what we're doing today isn't "your config sucks, so I'm going to deprogram Envoy", it's more like "your config sucks, but I'm still going to try to apply it, and Envoy might reject it." In my opinion, that's undefined behavior, which is worse.

Let's take this up in a separate issue or PR.

snorwin (approved):
@bittrance thanks a lot!

A lot was improved in this PR, really amazing. There's still room for some further improvements (see my comments) which we can follow up on in further PRs or issues.

/lgtm

PR #448, opened 2026-08-01
Populate GatewayClass supportedFeatures [GEP-2162]

bittrance:
Gateway API implementations are expected to announce what features they support, see https://gateway-api.sigs.k8s.io/geps/gep-2162/ . There are some obscure features which I don't fully understand, so this list is a bit of a guess. Enlightenment welcome.

### DIFF
--- a/pkg/gateway/features_test.go
+++ b/pkg/gateway/features_test.go
@@ -1,0 +1,67 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package gateway
+
+import (
+	"testing"
+
+	"sigs.k8s.io/gateway-api/pkg/features"
+)
+
+func TestSupportedFeaturesAreSorted(t *testing.T) {
+	for i := 1; i < len(supportedFeatures); i++ {
+		if supportedFeatures[i].Name < supportedFeatures[i-1].Name {
+			t.Errorf("supportedFeatures is not sorted: %q appears after %q",
+				supportedFeatures[i].Name, supportedFeatures[i-1].Name)
+		}
+	}
+}
+
+func TestSupportedFeaturesContainsCore(t *testing.T) {
+	required := []features.FeatureName{
+		features.SupportGateway,
+		features.SupportHTTPRoute,
+		features.SupportReferenceGrant,
+	}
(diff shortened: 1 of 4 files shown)


pkg/gateway/features_test.go
```diff
@@ -0,0 +1,67 @@
+/*
+Copyright 2026 The Kubernetes Authors.
```
snorwin, line 2:
```suggestion
Copyright The Kubernetes Authors.
```

@aojea in Gateway API we've stopped setting the year in the copyright notice as of this year, but I'm not sure how it's handled here?

pkg/gateway/features_test.go
```diff
@@ -0,0 +1,67 @@
...
+package gateway
+
+import (
+	"testing"
+
+	"sigs.k8s.io/gateway-api/pkg/features"
+)
+
+func TestSupportedFeaturesAreSorted(t *testing.T) {
+	for i := 1; i < len(supportedFeatures); i++ {
```
snorwin, line 26: I think the `"github.com/stretchr/testify/assert"` package  (e.g., for `IsIncreasing`, `Contains`) could really help with writing tests here. 

@aojea if you don't mind, could we add this as a dependency?

snorwin:
@bittrance now that Gateway API v1.6.1 is merged, could you rebase this PR so we can check the extended feature tests still pass?

pkg/gateway/features_test.go
aojea: we had this discussion in kubernetes multiple times and it is a philosophical thing, so I do not think there is good or bad, is a decision on how you want to maintain the code. In Kubernetes we have OWNERS per area so everyone can pick their philosophy, some SIGS prefer frameworks and other not, in all staging repo we try to avoid them because it increases the dependency graph, we do not have a problem like that here.

I'm in favor of the  golang idiomatic way of testing. I peronsally don't like high level frameworks. I do not think test code should be different the production code, having a high level API with a verb make the test easy to read semantically , but force you to depend on learning a new API that when you have races or weird behaviors  makes you have unnecessary problems , see https://github.com/kubernetes/kubernetes/issues/130346

I have to politely say no to this, the codebase here is very simple , and if possible I want the test to not hide the assertions

pkg/gateway/features_test.go
aojea: new guidelines from CNCF is to avoid the year but for practical reasons, is fine with or without it though

pkg/gateway/features_test.go
snorwin: That's fine with me, I was expecting this might be a controversial question.

pkg/gateway/features_test.go
bittrance: The linter complained about the year, so I removed it.

pkg/gateway/features_test.go
bittrance: Ya, the Golang world is split on this question. Aight, we continue in plain Go style.

snorwin (commented):
/lgtm

bittrance:
Rebase performed and conformance tests passes. However, I noticed that cpk will not reliably update `supportedFeatures` on a pre-existing GatewayClass. Will investigate.

aojea:
/approve

Thanks folks, appreciate all this bugfixes and features

PR #451, opened 2026-08-04
Only program Envoy if Gateway is Accepted=True

bittrance:
As per discussion in https://github.com/kubernetes-sigs/cloud-provider-kind/pull/447, this PR changes the logic to only program Envoy when the configuration is Accepted=True. Put differently, when encountering a bad configuration, we will retain the previous configuration.

It would be nice to have unit tests to cover this behavior, but that would require mocking both Envoy, the cache and various listers (some 200 lines in total if Claude's draft is to be believed), so I'm not sure it is worth it given that behavior is tightly tied to conditions that are tested by conformance tests?


aojea:
/ok-to-test

/approve

not an envoy expert so I trust your judgement
/assign @snorwin

aojea:
> It would be nice to have unit tests to cover this behavior, but that would require mocking both Envoy, the cache and various listers (some 200 lines in total if Claude's draft is to be believed), so I'm not sure it is worth it given that behavior is tightly tied to conditions that are tested by conformance tests?

@bittrance you can use a bats test https://github.com/kubernetes-sigs/cloud-provider-kind/tree/main/tests so you do not need to mock anything

bittrance:
@aojea Much more readable! Don't know how to get a stable test for an erroring envoy, but otherwise these tests capture the basic condition behavior.

Ready for review now.

snorwin:
@bittrance AFAIK Envoy rejects a configuration e.g., if it contains an invalid RE2 pattern. For the test, you could use a path match of type RegularExpression with an invalid expression.

bittrance:
@snorwin Will you look at that! https://github.com/kubernetes-sigs/cloud-provider-kind/issues/452 . This looks like it will break some new ground in `pkg/gateway/controller.go`, so I opened a separate issue for that problem.

I think the `xdsErr` we have today is in practice only about client errors such as communication failures; I'm not sure how to test for those.

aojea:
LGTM

@snorwin for final lgtm

tests/gateway_api.bats
```diff
@@ -0,0 +1,257 @@
...
+    [ "$output" = "InvalidParameters" ]
+
+    run kubectl get gateway test-gw-params \
+        -o 'jsonpath={.status.conditions[?(@.type=="Programmed")].status}'
+    [ "$output" = "False" ]
+
+    # Envoy must not be programmed: no address should be assigned.
+    run kubectl get gateway test-gw-params -o 'jsonpath={.status.addresses}'
+    [ -z "$output" ]
+
```
snorwin, line 55: @bittrance would it make sense to also assert that the container wasn't created here or would that bee too complex?

bittrance: @snorwin  Well, it is a negative test depending on implementation details (container labeling in this case), but why not? It is easy enough to check. I also added a positive container check to increase the chance of changes to those implementation details would be discovered. Fixed.

snorwin (approved):
@bittrance thanks. Really nice to see Gateway API support getting better step by step.

/lgtm

PR #453, opened 2026-08-07
fix: reflect static spec.addresses in Gateway status and enabel the corresponding conformance tests

snorwin:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it:**
cloud-provider-kind already supports `.spec.addresses` on Gateways, but the behavior does not match what the Gateway API conformance tests expect. This PR enables the `SupportGatewayStaticAddresses` feature and the corresponding conformance test, and fixes the status update to correctly set the `Accepted` and `Programmed` conditions when a requested address type is unsupported or otherwise unusable.

In addition, IPv6 is no longer enabled implicitly for Gateways that request only IPv4 addresses.

### DIFF
--- a/pkg/gateway/gateway.go
+++ b/pkg/gateway/gateway.go
@@ -6,10 +6,9 @@ import (
 	"fmt"
 	"net"
 	"reflect"
+	"slices"
 	"strings"
 	"time"
-
-	"google.golang.org/protobuf/types/known/durationpb"
 
 	clusterv3 "github.com/envoyproxy/go-control-plane/envoy/config/cluster/v3"
 	corev3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
@@ -20,6 +19,7 @@ import (
 	envoyproxytypes "github.com/envoyproxy/go-control-plane/pkg/cache/types"
 	cachev3 "github.com/envoyproxy/go-control-plane/pkg/cache/v3"
 	resourcev3 "github.com/envoyproxy/go-control-plane/pkg/resource/v3"
+	"google.golang.org/protobuf/types/known/durationpb"
 
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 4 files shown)


snorwin:
cc @bittrance

pkg/gateway/gateway.go
```diff
@@ -802,24 +786,32 @@ func createClusterLoadAssignment(clusterName, serviceHost string, servicePort ui
 	}
 }
 
-// getRouteHostnames determines the effective hostnames for a route.
-func getRouteHostnames(routeHostnames []gatewayv1.Hostname, listener gatewayv1.Listener) []string {
```
snorwin, line 806: note: `getRouteHostnames` was unused, so I removed it

pkg/gateway/gateway.go
```diff
@@ -857,41 +849,81 @@ func setAcceptedCondition(newGw *gatewayv1.Gateway) {
...
-		programmedCondition.Reason = string(gatewayv1.GatewayReasonInvalid)
-		programmedCondition.Message = "Gateway is not accepted"
+		// A Gateway can never be Programmed=True if Accepted=False.
+		meta.SetStatusCondition(&newGw.Status.Conditions, metav1.Condition{
+			Type:               string(gatewayv1.GatewayConditionProgrammed),
+			Status:             metav1.ConditionFalse,
+			Reason:             string(gatewayv1.GatewayReasonInvalid),
+			Message:            "Gateway is not accepted",
+			ObservedGeneration: newGw.Generation,
+		})
```
bittrance, line 861: I thought of making this change too. More readable with a full condition. 👍

bittrance (approved):
I think setAcceptedCondition and setProgrammedCondition no longer describe the role we give these functions. Perhaps evsluateXCondition? Other than that, looks good.

aojea:
@snorwin if you have the appetite, feel free to add an OWNERS file under `/pkg/gateway` and yourself as approver/reviewers , there is no commitment attached to it, but in case there are things like this you should not be waiting for me ;)

aojea:
/lgtm
/approve

Thanks

PR #456, opened 2026-08-18
gateway: implement GEP-1494 ExternalAuth HTTPRoute filter

aojea:
Add support for the experimental ExternalAuth HTTPRoute filter, which delegates authentication and authorization for a route rule to an external server speaking Envoy's ext_authz protocol over HTTP or GRPC.

The GEP-1494 feature names are only advertised on the GatewayClass when running the experimental release channel, where the CRD fields exist.

/assign @youngnick @snorwin @bittrance

### DIFF
--- a/pkg/gateway/externalauth.go
+++ b/pkg/gateway/externalauth.go
@@ -32,6 +32,7 @@ import (
 	"google.golang.org/protobuf/types/known/anypb"
 	"google.golang.org/protobuf/types/known/durationpb"
 
+	corev1 "k8s.io/api/core/v1"
 	corev1listers "k8s.io/client-go/listers/core/v1"
 	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
 	gatewaylistersv1 "sigs.k8s.io/gateway-api/pkg/client/listers/apis/v1"
@@ -128,14 +129,32 @@ func resolveExternalAuthBackend(
 		}
 	}
 
-	if _, err := serviceLister.Services(ns).Get(string(backendRef.Name)); err != nil {
+	svc, err := serviceLister.Services(ns).Get(string(backendRef.Name))
+	if err != nil {
 		return "", &ControllerError{
 			Reason:  string(gatewayv1.RouteReasonBackendNotFound),
 			Message: fmt.Sprintf("externalAuth backend Service %s/%s not found", ns, backendRef.Name),
 		}
 	}
+	port := int32(*backendRef.Port)
+	if !serviceHasPort(svc, port) {
+		return "", &ControllerError{
+			Reason:  string(gatewayv1.RouteReasonBackendNotFound),
+			Message: fmt.Sprintf("externalAuth backend Service %s/%s does not have port %d", ns, backendRef.Name, port),
+		}
+	}
 
-	return externalAuthClusterName(ns, string(backendRef.Name), int32(*backendRef.Port)), nil
+	return externalAuthClusterName(ns, string(backendRef.Name), port), nil
+}
+
+// serviceHasPort reports whether svc exposes the given Service port.
+func serviceHasPort(svc *corev1.Service, port int32) bool {
+	for _, p := range svc.Spec.Ports {
+		if p.Port == port {
+			return true
+		}
+	}
+	return false
 }
 
 // externalAuthClusterName returns the Envoy cluster name for an authorization server.
--- a/pkg/gateway/externalauth.go
+++ b/pkg/gateway/externalauth.go
@@ -1,0 +1,375 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package gateway
+
+import (
+	"crypto/sha256"
+	"fmt"
+	"time"
+
+	clusterv3 "github.com/envoyproxy/go-control-plane/envoy/config/cluster/v3"
+	mutationrulesv3 "github.com/envoyproxy/go-control-plane/envoy/config/common/mutation_rules/v3"
+	corev3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
+	extauthzv3 "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/ext_authz/v3"
+	hcm "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/network/http_connection_manager/v3"
+	upstreamsv3 "github.com/envoyproxy/go-control-plane/envoy/extensions/upstreams/http/v3"
+	matcherv3 "github.com/envoyproxy/go-control-plane/envoy/type/matcher/v3"
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/anypb"
+	"google.golang.org/protobuf/types/known/durationpb"
+
+	corev1listers "k8s.io/client-go/listers/core/v1"
+	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
+	gatewaylistersv1 "sigs.k8s.io/gateway-api/pkg/client/listers/apis/v1"
+)
+
+// GEP-1494 (HTTP Auth in Gateway API) is implemented on top of Envoy's ext_authz
+// filter. The authorization server is per HTTPRoute rule, but Envoy only allows
+// the server to be configured on the HTTP filter itself, so a distinct (disabled
+// by default) ext_authz filter is installed in the connection manager for every
+// unique ExternalAuth configuration, and each route enables the one it needs
+// through `typed_per_filter_config`.
+const (
+	// externalAuthFilterPrefix namespaces the generated Envoy HTTP filter names.
+	externalAuthFilterPrefix = "envoy.filters.http.ext_authz.gateway-api"
+	// externalAuthTimeout bounds how long a request waits for the authorization server.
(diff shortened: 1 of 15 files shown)


pkg/gateway/gateway.go
```diff
@@ -671,12 +688,12 @@ func ruleHasRedirectFilter(rule gatewayv1.HTTPRouteRule) bool {
 	return false
 }
 
-func (c *Controller) translateBackendRefToCluster(defaultNamespace string, backendRef gatewayv1.BackendRef) (*clusterv3.Cluster, error) {
+func translateBackendRefToCluster(serviceLister corev1listers.ServiceLister, defaultNamespace string, backendRef gatewayv1.BackendRef) (*clusterv3.Cluster, error) {
```
bittrance, line 691: What motivates evicting this func from the type?

pkg/gateway/httproute.go
```diff
@@ -25,24 +26,54 @@ func (e *noEffectiveBackendsError) Error() string {
...
-) (
-	envoyRoutes []*routev3.Route,
-	allValidBackendRefs []gatewayv1.BackendRef,
-	notAccepted *metav1.Condition,
-	resolvedRefsFailure *metav1.Condition,
-	partiallyInvalid *metav1.Condition,
-) {
-
+) routeTranslationResult {
+	var result routeTranslationResult
```
bittrance, line 76: 👍

tests/gateway-experimental/tests.bats
```diff
@@ -0,0 +1,162 @@
...
+
+@test "GatewayClass advertises the GEP-1494 features on the experimental channel" {
+    for i in {1..30}; do
+        FEATURES=$(kubectl get gatewayclass cloud-provider-kind \
+            -o 'jsonpath={.status.supportedFeatures[*].name}' 2>/dev/null)
+        [[ "$FEATURES" == *"HTTPRouteExternalAuth"* ]] && break
+        sleep 1
+    done
+
+    [[ "$FEATURES" == *"HTTPRouteExternalAuth"* ]]
```
bittrance: Substring of the other ones, so this check does not guarantee the base feature is present. I would use the jsonpath expression to look for them in turn.

pkg/gateway/httproute.go
```diff
@@ -118,20 +194,15 @@ func translateHTTPRouteToEnvoyRoutes(
 			}
 		}
 
-		buildRoutesForRule := func(match gatewayv1.HTTPRouteMatch, matchIndex int) {
```
bittrance, line 121: 👍 on getting rid of the inner function.

bittrance (commented):
A lot of code, but mostly straightforward. (It would have been helpful to have the refactoring steps as separate commits.)

Aight, I'll put up implementation during the weekend for the rest of my cases in kubernetes/gateway-api#5082 and we'll see if it passes.

pkg/gateway/gateway.go
aojea: it is reused now in other places like `func externalAuthCluster(serviceLister corev1listers.ServiceLister, routeNamespace, clusterName string, extAuth *gatewayv1.HTTPExternalAuthFilter) (*clusterv3.Cluster, error) {`

It can go either way, we keep extending the Controller methods or we plumb down the lister

pkg/gateway/httproute.go
aojea: yeah, too many outputs

pkg/gateway/gateway.go
bittrance: Sorry, bad search fu on my behalf. Can be resolved.

tests/gateway-experimental/tests.bats
aojea: good catch

aojea: fixed

bittrance (approved):

pkg/gateway/externalauth.go
```diff
@@ -0,0 +1,377 @@
...
+			Message: fmt.Sprintf("unsupported externalAuth protocol: %q", extAuth.ExternalAuthProtocol),
+		}
+	}
+
+	if extAuth.ForwardBody != nil && extAuth.ForwardBody.MaxSize > 0 {
+		extAuthz.WithRequestBody = &extauthzv3.BufferSettings{
+			MaxRequestBytes: uint32(extAuth.ForwardBody.MaxSize),
+			// The Gateway API requires bodies larger than maxSize to be
+			// truncated rather than rejected.
+			AllowPartialMessage: true,
```
bittrance: This says differently, I think? https://github.com/kubernetes-sigs/gateway-api/blob/b401cd4f7f5623cdbdf775898cea711c571c064e/apis/v1/httproute_types.go#L1690

aojea: There is other contradicting comment https://github.com/kubernetes-sigs/gateway-api/blob/b401cd4f7f5623cdbdf775898cea711c571c064e/apis/v1/httproute_types.go#L1788-L1802 or maybe I'm mixing things?

aojea: switched to false

bittrance:
I now have a first implementation of the full ExternalAuth conformance test suite (see https://github.com/kubernetes-sigs/gateway-api/issues/5082). FWIW, this implementation produces the same result as cilium 1.20.0, meaning that all tests pass except for two test cases:

1. when the ExternalAuth backendref is correctly configured, but there happens to be zero pods available to answer, both this implementation and cilium returns 403 (presumably Envoy default behavior) but I don't think this is a special case? If a normal backendref has zero targets, we return 5xx.
2. when the ExternalAuth backendref points an existing service, but on the wrong port, both this implementation and cilium returns 403. However, I think this is a known misconfiguration case (i.e. not one where we expect eventual consistency) and should return 5xx?

Neither of these are serious, but I think generating 5xx errors in these cases will make them easier to identify via metrics in some cases, for example when a REST API has a single client.

Tests are at https://github.com/bittrance/gateway-api/compare/main...bittrance:gateway-api:conformance-external-auth-part-2?expand=1 .

youngnick:
@bittrance, I think I can see where Envoy is coming from on the first use case though, as if the thing that's doing the authentication and/or authorization is unavailable, then Envoy is failing closed and rejecting the request as unauthorized.

I agree that the second is a misconfiguration, so it does make sense that it might be a 500-class error.

That said, I suspect that changing these will require Envoy changes (and probably other dataplane in other proxies where it doesn't work), so maybe we should consider setting the test to just say "must fail in some way, aka must be a 4xx or 5xx" to begin with?

It's way easier to tighten the tests up later that it is to expand them later.

Edit: To be clear, I'm talking about the upstream tests, not the ones on this impl, sorry.

bittrance:
@youngnick Actually, Regarding case 1, Envoy has [status_on_error](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_authz/v3/ext_authz.proto#envoy-v3-api-msg-extensions-filters-http-ext-authz-v3-extauthz) for this purpose. However, I can see that if you are a serious SaaS you don't want to leak needless implementation details, ever. I will keep the wider set of accepted statuses in the externalauth conformance tests for this case.

@aojea I think cloud-provider-kind should set this to 502 as its intended audience are developers that will want to know that something is off. What do you think?

aojea:
> @aojea I think cloud-provider-kind should set this to 502 as its intended audience are developers that will want to know that something is off. What do you think?

cloud-provider-kind audience is developers but the goal is to provide production behaviors to local development, so whatever we decide to put as an error is the one we expect implementations to use it ... 

I lean towards keeping this a 403. The client making the service request shouldn't need to know how the auth chain is configured, the end result from their perspective is simply that they are not authorized.

@youngnick what are you thinking for implementing in cilium? @robscott how are we handling these misconfiguration of external auth internally?

youngnick:
I don't remember what Cilium does, but I would guess we'll do Envoy's default behavior if we don't already. I think that 403 is generally the right case, but again, I don't think the upstream conformance tests need to be picky there, as long as the request fails.

aojea:
ok, it seems we get approval from @bittrance so this is ready for merging once some gateway api folk LGTM it 

@youngnick @snorwin , since this is mainly for developing and testing environment is almost free to iterate on experimental features

bittrance:
@aojea  PR https://github.com/kubernetes-sigs/gateway-api/pull/5113 now has the full suite of conformance tests. I have widened the list of accepted "misconfiguration" statuses to include 403 (and 503). That means that according to that suite, this PR fails two tests:

- a backendref with misconfigured extauth service port should say ResolvedRefs=False
- having one correct and one invalid extauth service should trigger PartiallyInvalid

Apart from that, I think we are fine.

aojea:
> * a backendref with misconfigured extauth service port should say ResolvedRefs=False
> * having one correct and one invalid extauth service should trigger PartiallyInvalid
> 
> Apart from that, I think we are fine.

ok, I need to fix that

aojea:
@bittrance pushed and tested with your PR and now is passing, if you are ok please LGTM here ... we can always iterate if the definition of conformance change but I like to have a starting point and your PR seems a solid one to me

bittrance:
/lgtm
?

aojea:
> @bittrance: changing LGTM is restricted to collaborators
> 
> Details

I see, you are not a member, if you plan to keep contributing to gateway api and here I'm happy to sponsor you to become a member https://github.com/kubernetes/community/blob/main/community-membership.md#member , you just need a second sponsor

stmcginnis:
> > @bittrance: changing LGTM is restricted to collaborators
> > Details
> 
> I see, you are not a member, if you plan to keep contributing to gateway api and here I'm happy to sponsor you to become a member https://github.com/kubernetes/community/blob/main/community-membership.md#member , you just need a second sponsor

Feel free to tag me for a second sponsor. I've noticed some of your contributions and am happy to support you becoming a member.

youngnick:
I only took a brief look at the code, but from what I see, LGTM. Thanks as always @aojea!

aojea:
I need a "real" lgtm to merge, @bittrance now you have the powers to do it

bittrance:
/lgtm
!

aojea:
> /lgtm !

new CI systems  are always surprising, in Kubernetes we use prow that is github comment driven

PR #463, opened 2026-08-24
feat: add GRPCRoute support

squizzi:
- Translate `GRPCRoute` into Envoy routes (method and header matches, weighted backends, `ReferenceGrant` checks, and route status).
- Enable `GRPCRoute` as a supported kind on HTTP/HTTPS listeners.
- Negotiate HTTP/2 for gRPC: ALPN `h2`/`http/1.1` on HTTPS listeners, and HTTP/2 protocol options on gRPC upstream clusters.


aojea:
please try to make it additive, in the sense that minimize the diff with the existing code and avoid drive by refactors

pkg/gateway/gateway.go
```diff
@@ -184,6 +187,28 @@ func (c *Controller) buildEnvoyResourcesForGateway(gateway *gatewayv1.Gateway) (
...
+		if len(acceptingListeners) > 0 {
+			processedListeners := make(map[gatewayv1.SectionName]bool)
+			for _, listener := range acceptingListeners {
+				if _, ok := processedListeners[listener.Name]; !ok {
+					grpcRoutesByListener[listener.Name] = append(grpcRoutesByListener[listener.Name], grpcRoute)
+					processedListeners[listener.Name] = true
+				}
+			}
+		}
+	}
```
bittrance, line 285: The two validation-and-transfer loops are effectively identical and it is relevant that they stay identical. Could we refactor them either into one loop or into a helper?

pkg/gateway/grpcroute.go
```diff
@@ -0,0 +1,692 @@
...
+			to := gatewayv1.ReferenceGrantTo{
+				Group: "",
+				Kind:  "Service",
+				Name:  &backendRef.Name,
+			}
+			if !isCrossNamespaceRefAllowed(from, to, ns, referenceGrantLister) {
+				if firstErr == nil {
+					firstErr = &ControllerError{
+						Reason:  string(gatewayv1.RouteReasonRefNotPermitted),
+						Message: "permission error",
```
bittrance: The main audience are developers, so messages should be informative. For example, in #456 a similar error says `fmt.Sprintf("reference to Service %s/%s not permitted by any ReferenceGrant", ns, backendRef.Name),`

pkg/gateway/grpcroute.go
```diff
@@ -0,0 +1,692 @@
...
+					}
+				}
+				continue
+			}
+		}
+		if _, err := serviceLister.Services(ns).Get(string(backendRef.Name)); err != nil {
+			if firstErr == nil {
+				firstErr = &ControllerError{
+					Reason:  string(gatewayv1.RouteReasonBackendNotFound),
+					Message: "backend not found",
```
bittrance: Same comment about message here: in a route with multiple mirrors or extauths, it would help to know which backend was not found.

bittrance (commented):
I can see this is still draft, but here are some random things I noted when taking a quick look.

pkg/gateway/grpcroute.go
squizzi: ya that's fair, I think I'm habitually keeping errors secretive when this a dev tool, noted -- thanks.

PR #126, opened 2024-08-30
support different platforms and connectivity scenarios

aojea:
These are these combinations

- Two type of platforms: containers reachable (Linux) or not reachable from host (Mac, Windows they run in a VM )
- Two type of deployments: cloud-provider-kind as a container or as a proces.

1. Containers Reachable and KCCM Process: Direct (Linux) (autodetected)
2. Containers Reachable and KCCM container: Direct (Linux) (autodetected)
3. Containers Not Reachable and KCCM Process: Tunnel (Mac, Windows, WSL2) (autodetected)
4. Containers Not Reachable and KCCM Container: Portmap (Mac, Windows, WSL2) (set via flag)

Fixes: #109

PR #466, opened 2026-08-27
feat: add --enable-lb-tunnel option for linux

sshambar:
Adds a linux-only command option '--enable-lb-tunnel' to enable Darwin/windows default behavior. 
Requires sudo (same as Darwin), and the environment should be setup to allow connection the kind cluster (eg CONTAINER_HOST for non-root podman).

The load balancer IPs are added to the loopback interface just as for the Darwin build.


aojea:
@sshambar just curios, when do you need this option in linux?

sshambar:
> @sshambar just curios, when do you need this option in linux?

When I'm running kind in rootless pods (rootless podman in my case), the load manager ip/ports aren't directly reachable on the local machine, and I need to discover/map the actual published port.

I found when running cloud-provider-kind on Darwin (also rootless podman), if a service's External-IP is 10.89.0.12 on port 53, I could just connect to that.  On linux, that wasn't true.  Adding the option allows that.

It's nice that once the external-ip is available on the host, I can create a subdomain (<domain>) using the k8s_external coredns plugin, create a LB service for coredns, point the local resolver at it's external-ip for <domain>, and then any LB service is available directly on the machine via <name>.<namespace>.<domain> on it's published port.  Just like bare-metal k8s but with containers safely locked in user-mode :)

I've done something similar with a script to link veth peers to the podman infra network, and add a route to the network - which is nice as it offers native speed - but it's a podman+linux only solution (github.com/sshambar/setup-rootless/ if you're curious).  I also played with doing something similar on MacOS with vmnet (https://github.com/nirs/vmnet-helper) but it requires managing my own podman machine instance...

aojea:
/ok-to-test

It's ok, I didn't understand where this scenario was needed,  you please validate this option solves your issue, I can not test it

cmd/app.go
```diff
@@ -164,6 +175,8 @@ func runE(cmd *cobra.Command, args []string) error {
 	// some platforms require to enable tunneling for the LoadBalancers
 	if runtime.GOOS == "darwin" || runtime.GOOS == "windows" || isWSL2() {
 		config.DefaultConfig.LoadBalancerConnectivity = config.Tunnel
+	} else if enableLBTunnel {
```
aojea: is there any heuristic we can be smart and avoid another flag?

sshambar: Well, linux running as euid 0 might be good enough, but I wasn't sure if people used cloud-provider-kind that way now and expected it to continue to "not-publish".  Combined with checking if the attached docker/podman instance was rootless would be safer, but I'm not sure how many assumptions we'd have to make to just avoid a flag...

sshambar:
> It's ok, I didn't understand where this scenario was needed, you please validate this option solves your issue, I can not test it

Yes, I tested it on rootless podman, and it works as expected.  I could try rootless docker, I'd just need to install it somewhere :)

aojea:
> It's nice that once the external-ip is available on the host, I can create a subdomain () using the k8s_external coredns plugin, create a LB service for coredns, point the local resolver at it's external-ip for , and then any LB service is available directly on the machine via .. on it's published port. Just like bare-metal k8s but with containers safely locked in user-mode :)

ok, so this is about exposing the coreDNS running in the cluster?

because there are simpler methods than this to achieve that

sshambar:
> ok, so this is about exposing the coreDNS running in the cluster?

No, this is about exposing the load balancer ips on the loopback interface :)

aojea:
please bear with me since I'm misunderstanding how this podman rootless work

> When I'm running kind in rootless pods (rootless podman in my case), the load manager ip/ports aren't directly reachable on the local machine, and I need to discover/map the actual published port.

why are not reachable? are those running inside a vm or something?

sshambar:
> please bear with me since I'm misunderstanding how this podman rootless work
> why are not reachable? are those running inside a vm or something?

On linux, rootless podman runs the kind control-plane/worker nodes and the cloud-provider-kind envoy proxies in the pod network "kind" -- which is in a different network namespace than the host (let's say the subnet assigned to it is 10.89.0.0/24).  In handling a LoadBalancer service, cloud-provider-kind creates the envoy container (eg at ip 10.89.0.12) and load balances the original service port (eg 443) to the control-plane/worker node ports (also on the 10.89.0.0/24 network, but at the NodePort on that node).

On Darwin, since the containers all run in a VM (a "podman machine"), cloud-provider-kind assigns the envoy ip (eg 10.89.0.12) to the host loopback interface, and forwards connections to the service ports into the pod network.  ie. connections on the host to 10.89.0.12:443 work.

However, on linux it doesn't do that, so the root network namespace has no route to the pod network (connections to 10.89.0.12:443 fail).

The new option just requests cloud-provider-kind assign the address to the loopback interface and forward connections to the pod network namespace, just as it does on Darwin.  That way connections on the host (eg 10.89.0.12:443) connect just as they do on Darwin.

Hope that explains the issue clearly enough; let me know if you still have questions :)

aojea:
> On linux, rootless podman runs the kind control-plane/worker nodes and the cloud-provider-kind envoy proxies in the pod network "kind

but are they bundling this? ... if this is about an integration we can talk with them, I worked at RH before and those are great guys, I rather prefer to create a nice end to end user story than trying to start implementing workarounds ... we may end with this solution, but is better if we think more holistically about this

sshambar:
> but are they bundling this? 

Bundling?  Rootless mode is just running the container manager as a non-privileged user... docker does the same thing.

To verify that the patch works for docker I setup kind in rootless docker, and indeed the kind nodes are in a namespaced network and none of the assigned IPs are addressable on the host.

Running the patched cloud-provider-kind --enable-lb-tunnel (w/ sudo, and DOCKER_HOST set) exposes the IP's on the loopback and the ports are directly addressable.

So the patch does make the cloud-provider work just as it does on Darwin (and I guess WSL2), both both rootless podman and docker.

aojea:
> To verify that the patch works for docker I setup kind in rootless docker, and indeed the kind nodes are in a namespaced network and none of the assigned IPs are addressable on the host.

this is the part I'm unfamiliar, I do not know how networking works on these rootless platforms https://github.com/podman-container-tools/podman/blob/main/docs/tutorials/basic_networking.md#pasta \

I'd like  to know if we can avoid adding flags

sshambar:
> I'd like to know if we can avoid adding flags

The only way I could see to avoid the flag is making assumptions based on the environment... for example cloud-provider-kind could (when run as EUID 0) determine if the container manager is running rootless, which would "imply" that the provider was being run privileged strictly to expose the LB IPs publicly (the way a real cloud provider would)... and switch to tunnel mode.  It'd be a noop on Darwin (always rootless), and probably on WSL2 (wish I had an intel machine to test it on), but the special case on Linux since normally cloud-provider-kind would be running as root with a rootful provider, or non-root with a rootless one.

So if running as root, and the provider isn't, assume tunneling is being requested.

I know how to detect rootless on podman and docker, but I don't know much about the other providers (nerdctl, finch, lima).

aojea:
Thanks for exploring all the options, I think is clear that is not simple and the patch is small and seems well contained behind a lot of assertions on runtimes and system properties, @sshambar is this working fine for you , have you tested it? can you add some note on the readme or in the docs to briefly explain this so next rootless people can figure this out https://github.com/kubernetes-sigs/cloud-provider-kind/tree/main/docs

sshambar:
>  can you add some note on the readme or in the docs to briefly explain this so next rootless people can figure this out 

Sure, I'll work on referencing it in the docs... 

If you like, I could take a crack at rootless detection (so that "by default" it works for rootless users if correctly run as sudo)... and perhaps leave a "reverse" flag to disable the auto-detection if it's really undesired for some reason?

sshambar:
... as the "reverse" flag would effectively be a "disable tunnel" flag, it might prove useful for anyone who'd like to run cloud-provider-kind (as a cloud-provider only, not a forwarder) on Mac without sudo -- so with the flag on Mac it would behave exactly as the current release does on rootless Linux.  Currently, the app requires sudo to run at all on the Mac, so it'd just a nice side-benefit :)

sshambar:
I have a working build that detects rootless on docker, podman and nerdctl (all Linux).  I could submit an alternate PR with that patch if you'd like to investigate it.  A more general patch might consider "remote" detection... as that would correctly detect Mac and Windows environments for tunneling exactly as the current builds but would also handle remote containers on Linux (ie Mac/windows are already "remote" in that they're on VMs, and that would default to tunnels on Linux when connected to a VM on that platform too...)

sshambar:
OK, I've been testing the rootless detection, and nerdctl appears to have difficulties reporting rootless mode correctly when run with sudo but the environment directing it to the rootless containerd/rootlesskit.  So, the detection can't be completely automatic (even if it could default correctly under docker/podman)... so I could add rootless detection but it still needs the command arg.

I could adapt the command arg to support using it to set the tunnel true or false overriding the auto-detection if present... which would still allow platforms like MacOS to opt-out of the sudo/tunnel if desired?

aojea:
> I could adapt the command arg to support using it to set the tunnel true or false overriding the auto-detection if present... which would still allow platforms like MacOS to opt-out of the sudo/tunnel if desired?

I prefer to do changes when there is more evidance ... is MacOs rootless a thing?

cmd/app.go
```diff
@@ -95,6 +101,11 @@ func runE(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("please run this again with `sudo`")
 	}
 
+	// Process on linux with tunnel must run using sudo
+	if runtime.GOOS == "linux" && enableLBTunnel && syscall.Geteuid() != 0 {
+		return fmt.Errorf("please run this again with `sudo`")
+	}
+
```
aojea: do we need to fail if both `enableLBPortMapping` and `enableLBTunnel` are configured at the same time?

sshambar: > do we need to fail if both `enableLBPortMapping` and `enableLBTunnel` are configured at the same time?

Well, my read of tunnel mode is that it requires port mapping to work, so it's kinda redundant.

sshambar:
> I prefer to do changes when there is more evidance ... is MacOs rootless a thing?

Well... podman machine on mac can setup rootless or roolful modes as a choice... but it doesn't really matter since both modes are "remote" and therefore only work in tunnel mode.

sshambar:
BTW, I'm still playing with the rootless auto-detect mode on linux.  I can't seem to get around needing the --enable-lb-tunnel flag though, since I can't find a simple way to correctly auto-detect rootless on nerdctl -- and there needs to be some "I know what I'm doing" flag to enable tunnel mode for nerdctl when cloud-provider-kind is running as EUID 0...

Would you like me to update the PR with the auto-detect logic I have so far?

cmd/app.go
aojea: see lines 179 to 185

```go

		config.DefaultConfig.LoadBalancerConnectivity = config.Tunnel
	}

	// flag overrides autodetection
	if enableLBPortMapping {
		config.DefaultConfig.LoadBalancerConnectivity = config.Portmap
	}

```

one is going to override the other

aojea:
> and there needs to be some "I know what I'm doing" flag to enable tunnel mode for nerdctl when cloud-provider-kind is running as EUID 0...

will not be that podman and docker gains new better behavior and nerdctl remains as it is today?

If that is the case is a net positive case, once some nerdctl person hits it we reconsider

cmd/app.go
sshambar: In my updated patch with rootless auto-detection, I changed the logic so either LB flag overrides all LB auto-detection... I guess I should just push the new code and keep the current patch on a branch which I can restore if you don't want to keep the auto-detection logic.  I'll force push the detection code so you can review it :)

sshambar:
BTW, the updated code flow is as follows:
 - using --enable-lb-tunnel=(true|false) or --enable-lb-port-mapping=(true|false) override LB auto-detection
 - LB auto-detection needs to happen after the container provider has been determined/detected (so LB auto-detection is moved after that)
 - Linux rootless auto-detection works for docker/podman, and is present for nerdctl (but fails to show rootless in it's current nerdctl release, may be fixed in a later release though)
 - MacOS, windows and WSL2 LB detection the same as before, but can now be overridden by the --enable-lb flags
 - The MacOS/Linux tunnel mode sudo check is done after LB mode is determined (manual or auto)

So the end result (without flags):
 a) Same for non-linux platforms and linux as EUID != 0.
 b) Linux as EUID 0 detects rootless, and enables tunnel mode.

Any --enable-lb-* flags (true or false) override all LB auto-detection -- and permit "direct" LB mode on MacOS/windows which wasn't an option before (but requires --enable-lb-tunnel=false)

Tunnel mode requested or detected still requires EUID 0.

Ran through a bunch of tests in different environments, and all the logic should be the same as before.  All new logic confined to "linux EUID 0 connect to rootless", or requires using the new --enable-lb-tunnel flag.

aojea:
linter issues ... try to explain that in a comment on top of the current logic since it is getting complicated and will be impossible to remember the context and reasons just by looking at the code

sshambar:
> linter issues ...

Fixed all but the switch lint, as if-else chain is simpler with overriding booleans

> try to explain that in a comment on top of the current logic since it is getting complicated and will be impossible to remember the context and reasons just by looking at the code

I added comments throughout the logic to explain what's going on... I could add more detail if you think it's not clear enough. I also added better logging when auto-detection is used.

I exposed --enable-lb-tunnel on MacOS/WSL so that cloud-provider-kind may be run unprivileged on those platforms with `--enable-lb-tunnel=false` -- it's similar to using `--enable-lb-port-mapping` on those platforms, but doesn't publish the ports which useful for keeping the cloud-provider isolated in the container network (great for testing/security!)

I plan on including a description of the new option in the documentation too, but I wanted to wait until the final logic is settled :)

sshambar:
BTW, I think I have a workaround for nerdctl (which would also work for docker/podman)... if cloud-provider-kind is run with CAP_NET_ADMIN it should still support tunnel mode (and would be more secure).  Perhaps the feedback about running under sudo would include suggestions to attach the capabilities to the file with sudo, and then run it as the rootless user?

aojea:
kind and cloud provider kind audience are not power users and we like to keep a minimum surface, sane defaults and avoid to get too much knobs that can end in complex configurations that can end in "shoot in the feet" kind situations.

See here the number of complex network configurations exist https://github.com/kubernetes-sigs/cloud-provider-kind/pull/126

So I prefer to avoid offering too much flexibility, I think the userspace case is legit but I never heard of macs or windows wanting rootless, is there any benefit on that? I prefer if we solve just the problem at hand instead of growing scope

sshambar:
> kind and cloud provider kind audience are not power users and we like to keep a minimum surface, sane defaults and avoid to get too much knobs that can end in complex configurations that can end in "shoot in the feet" kind situations.

That's fair... I just figured that since the flag is needed on linux, it might be nice to apply it universally.  But I can revert it to linux-only if you prefer.

> So I prefer to avoid offering too much flexibility, I think the userspace case is legit but I never heard of macs or windows wanting rootless, is there any benefit on that?

On mac/windows, rootless or not isn't really a factor since all the container managers are "remote" (in a VM)... so the rootless detection is only for linux.  Using the "lb-tunnel=false" flag on those platforms would just be to turn off the port mapping in the ccm containers...

BTW, I tried using the capability workaround for nerdctl, but nerdctl has other problems with rootless (like not supporting --publish-all or dynamically allocating ports) so there's really no reason to work around it's rootless detection (basically, it's rootless support is not yet docker compatible).

The capability workaround works great on podman/docker tough... so I could add a check for the Ambient network capabilites on linux rather than full EUID = 0 -- but that's something I could also do in a followup PR if you'd prefer :)

sshambar:
I've updated the PR so --enable-lb-tunnel is linux only, and updated the README/docs to discuss linux rootless support.  Let me know if you think further clarification is needed.

cmd/app.go
```diff
@@ -214,6 +211,40 @@ func runE(cmd *cobra.Command, args []string) error {
 			option = cluster.ProviderWithDocker()
 		}
 	}
+
+	// any --enable-lb-* flags (true or false) override lb-tunnel auto-detection.
+	//nolint:gocritic // if-else chain is more readable than switch
```
aojea, line 216: haha oh man you like to start a war XD (I do not have strong opinion but this topic is always fun)

cmd/app.go
```diff
@@ -214,6 +211,40 @@ func runE(cmd *cobra.Command, args []string) error {
...
+		// auto-detect lb-tunnel
+		if (runtime.GOOS == "linux" && syscall.Geteuid() == 0) {
+			// linux run as sudo
+			if isRootless, err := container.DetectRootless(); err != nil {
+				klog.Infof("Failed to auto-detect lb-tunnel mode: %s", err)
+			} else if isRootless {
+				// if containers are rootless, choose tunnel mode
+				klog.Infof("Containers are rootless; using lb-tunnel mode")
+				config.DefaultConfig.LoadBalancerConnectivity = config.Tunnel
+			}
```
aojea, line 231: one clarification, for this to work cpk need also to be running ON the host, some people run it inside a container ... the runtime detect allows us to differentiate between running inside the host and running in a container ... don't we need to discriminate here somehow?

cmd/app.go
```diff
@@ -214,6 +211,40 @@ func runE(cmd *cobra.Command, args []string) error {
 			option = cluster.ProviderWithDocker()
 		}
 	}
+
+	// any --enable-lb-* flags (true or false) override lb-tunnel auto-detection.
```
aojea, line 215: this is this logic, right?
```
var disableAutoDetect bool
	cmd.Flags().VisitAll(func(flag *pflag.Flag) {
		klog.Infof("FLAG: --%s=%q", flag.Name, flag.Value)
		// presence of these flags override auto-detection
		if flag.Name == "enable-lb-port-mapping" || flag.Name == "enable-lb-tunnel" {
			if flag.Changed {
				disableAutoDetect = true
			}
		}
	})
```

flags are always tricky, is this `flag.Changed` reliably and does not have weird cases with the defaults?

cmd/app.go
sshambar: Yes, both docker and podman report on the rootless-ness of the target container manager (otherwise running as sudo with the DOCKER_HOST/CONTAINER_HOST wouldn't work).  nerdctl only reported on the client's rootlessness which is why it doesn't detect correctly -- in fact I should really remove it from the rootless detection switch as it will never return a positive result, and they call `sudo nerdctl` "misusage" (ref https://github.com/containerd/nerdctl/issues/5178)

The tricky part about running cpk containerized is of course the `sudo docker run --network host --privileged...` part runs in the root dockerd (so the image must be loaded there) -- but once the socket is mounted in the container the containerized docker correctly connects to rootless dockerd and everything works as expected.

Do you think I should add a section in the rootless docs describing how to correctly use sudo for containerized?

cmd/app.go
sshambar: I agree it's a "poorly named" flag, but in reality it's a "presence" flag, not a "changed" flag... if the flag is present and matches the default, it is still "true" -- api doc: "If the user set the value (or if left to default)".  Loads of programs depend on that behavior and it would be an api break to change how it works.  (it could also be accomplished with a function to handle the flag, but that's just code duplication :)

cmd/app.go
aojea: I always ask @AkihiroSuda for these topics, IIRC he is the nerdctl creator and one of the most relevant persons in the container rootless ecosystem

AkihiroSuda: I don't get the context.
Why do you want to run `sudo nerdctl` against a rootless socket?

> but once the socket is mounted in the container the containerized docker correctly connects to rootless dockerd and everything works as expected.

Bind-mounting the socket is not supported in the case of nerdctl

sshambar: @AkihiroSuda
> I don't get the context. Why do you want to run `sudo nerdctl` against a rootless socket?

cloud-provider-kind only requires sudo when run in tunnel mode (so it can add IP addrs to loopback, and listen on privileged sockets on them).  However, as root it still needs to create containers in the container manager for the user who ran it as sudo.  The way it connects to the container manager is via a socket.  docker (DOCKER_HOST) and podman (CONTAINER_CONNECTION) both support remote access this way.  However, I couldn't get nerdctl to work with just CONTAINERD_ADDRESS.

> Bind-mounting the socket is not supported in the case of nerdctl

Bind-mounting is a special case; it's only required when running cloud-provider-kind itself inside a container.

sshambar: > I don't get the context. Why do you want to run `sudo nerdctl` against a rootless socket?

BTW, does containerd support a network context?  docker and podman use this on Mac/WSL to connect to their container managers in VMs and it's how sudo cloud-provider-kind runs on those platforms... perhaps we could use something similar with nerdctl?

AkihiroSuda: What is network context?

AkihiroSuda: > However, I couldn't get nerdctl to work with just CONTAINERD_ADDRESS.

You can just stat the `$CONTAINERD_ADDRESS` socket and check its owner UID

sshambar: > What is network context?

Sorry for the delay (busy weekend :)
Basically it's connection over network... `docker context` (or podman) let you use ssh:// or tcp:// connections to communicate with remote instances (VM, another host, but works locally between users too).  I was thinking if nerdctl supported that too, it could work as an alternative to the unix-socket interface.

sshambar: > You can just stat the `$CONTAINERD_ADDRESS` socket and check its owner UID

That could be one method to determine if containerd was running non-root... but doesn't really work for cloud-provider-kind  as `nerdctl run ...` as root doesn't work against the socket of another user (with both docker and podman it does)... which is why I was suggesting using a network context (see above).

AkihiroSuda: >  I was thinking if nerdctl supported that too, it could work as an alternative to the unix-socket interface.

Unsupported, as the daemon and the client need to share the filesystem currently.

aojea:
@sshambar sorry for the delay, catching up, are we good with current state or are there some pending issues?

sshambar:
> @sshambar sorry for the delay, catching up, are we good with current state or are there some pending issues?

I'm thinking the only thing I should change is removing an attempt to special case nerdctl for rootless detection since our discussions have made it clear that it doesn't support access from another (ie root) user (it would require running cloud-provider-kind in a split-permissions mode, and is really beyond the current goal :)

sshambar:
In addition, I may add instructions on how to run cloud-provider-kind with sudo so that it correctly inherits the user's local environment (which requires setting the correct environment variables, or configuring the root user's environment)

aojea:
ok, tag me once you are happy with the state, as we discussed, try to keep it focused to the problem at hand and avoid overgeneralizing

sshambar:
@aojea ok I removed the nerdctl test, and added some text on rootless_support.md explaining the sudo environment to retain rootless connectivity.   Let me know if you think if it isn't clear enough...

cmd/app.go
```diff
@@ -86,15 +90,17 @@ func newVersionCommand() *cobra.Command {
 
 func runE(cmd *cobra.Command, args []string) error {
 	// Log flags
+	var disableAutoDetect bool
 	cmd.Flags().VisitAll(func(flag *pflag.Flag) {
 		klog.Infof("FLAG: --%s=%q", flag.Name, flag.Value)
+		// presence of these flags override auto-detection
+		if flag.Name == "enable-lb-port-mapping" || flag.Name == "enable-lb-tunnel" {
+			if flag.Changed {
+				disableAutoDetect = true
```
aojea: `flag.Changed` is now true for `--enable-lb-port-mapping=false`. On Mac/Windows/WSL2 that was a no-op (tunnel was still auto-selected); now it skips autodetection and the process runs in direct mode. I think that only if `--enable-lb-tunnel=false` should disable autodetection,

cmd/app.go
```diff
@@ -214,6 +211,40 @@ func runE(cmd *cobra.Command, args []string) error {
...
+		if runtime.GOOS == "darwin" || runtime.GOOS == "windows" || isWSL2() {
+			klog.Infof("Containers are remote; using lb-tunnel mode")
+			config.DefaultConfig.LoadBalancerConnectivity = config.Tunnel
+		}
+	}
+
+	if config.DefaultConfig.LoadBalancerConnectivity == config.Tunnel {
+		// Tunneling on macOS/linux requires elevated privilege
+		if (runtime.GOOS == "darwin" || runtime.GOOS == "linux") && syscall.Geteuid() != 0 {
+			return fmt.Errorf("please run this again with `sudo` or --enable-lb-tunnel=false")
```
aojea: > -enable-lb-tunnel=false

this only works in linux so the message should be different per platform

pkg/container/container.go
```diff
@@ -48,6 +48,26 @@ func SetRuntime(name string) {
...
+	}
+	if fstr == "" {
+		return false, fmt.Errorf("unable to determine rootless for provider %s, assuming false", containerRuntime)
+	}
+	cmd := kindexec.Command(containerRuntime, "info", "-f", fstr)
+	result, err := kindexec.OutputLines(cmd)
+	if err != nil {
+		return false, err
+	}
+	return result[0] == "true", nil
```
aojea: be defensive and check that len(result0 > 0 to avoid panics

aojea:
One more case, what happens when cloud-provider-kind runs in a container against a rootless daemon? (I think I already asked but can not find it)

cmd/app.go
sshambar: That was an intentional decision to enable the ability to disable port-mapping, which is otherwise not possible on Mac/Windows.  I've had a number of situations on MacOS where port-mapping failed (duplicate port #s) and it was nice to have the ability to turn the feature off.  I could restore the original behavior and submit another PR (?), but it makes the code more complex...

cmd/app.go
sshambar: Good catch, that was leftover from when I'd unified the flags for all platforms... I'll fix it to reference lb-port-mapping on Darwin instead (CPK works quite well on Mac w/o sudo with the flag 'false', it just can't add external IPs)

pkg/container/container.go
sshambar: yep, better safe... :)

sshambar:
> One more case, what happens when cloud-provider-kind runs in a container against a rootless daemon? (I think I already asked but can not find it)

Yeah, we covered that back in https://github.com/kubernetes-sigs/cloud-provider-kind/pull/466#discussion_r3936366653

cmd/app.go
aojea: interesting, if it does not adds external ips how do you use it?

cmd/app.go
aojea: it is ok, but I'm still curious about how do you use it without tunnel or port-mapping

sshambar: > it is ok, but I'm still curious about how do you use it without tunnel or port-mapping

It's really useful when I have workloads that wait for LoadBalancer services to be assigned before allowing further service activation... that way I can test apps in isolated test environments that normally would only work in cloud environments (all without the need for sudo :)

aojea:
linter error, then we can move on, there are many flags and combinations at one point but this does not look like change anything that exist today
LGTM once you fix the linter

PR #475, opened 2026-09-18
loadbalancer: support upstream proxy protocol

hown3d:
To preserve client IPs e.g. when testing out access restrictions in upstream proxies, it's required for the upstream proxy to retrieve the original client IP.

This PR introduces a new annotation for services that can be used to configure proxy protocol.


aojea:
can you speak more about the feature request , what motivates it, how are you using it, ... ? so we try to put into the wider kubernetes ecosystem and see if we can find some common themes?

The goal of the project is to offer an standard/reference implementation, if we start to add annotations we become yet another different thing ...

hown3d:
We are trying to create a local development environment for a controller that configures IP-based whitelisting on different proxies, e.g istio and Envoy Gateway. 

In order to correctly test if whitelisting works as expected, we must be able to differentiate between clients in these proxies by IP.

To do so, we want to expose these proxies using the ccm's LoadBalancers with proxy protocol to preserve the client IP.

aojea:
@hown3d I sympathize with the problem but I really really want to avoid to grow scope based on annotations, as this will make it harder to say no to other things and we end in a similar situation as ingress-nginx ... please check this PR  https://github.com/kubernetes-sigs/cloud-provider-kind/pull/477 , it allows power users to implement their custom behaviors based on annotations without impacting the stability of the project

